PR #33, opened 2016-02-04
added 178 websites that got blocked

RajkumarPandi:


hellais:
Hi @RajkumarPandi thanks for suggesting new URLs to be added to the testing list for india. However in order for us to review and potentially integrate this into the repository of testing you will have to format them as `csv` to include also the category codes for each URL.

Also what is the source of this list of blocked websites? If the source is something authoritative (such a government issued blocklist) and we can reference a document you may consider adding it to the official lists that are found here: https://github.com/citizenlab/test-lists/tree/master/lists/official/.

For reference you can see how the government issued blocklists for italy look like: https://github.com/citizenlab/test-lists/tree/master/lists/official/it.

Ideally there would also be a parser and downloader to update the lists added here: https://github.com/citizenlab/test-lists/tree/master/lib/lists/official

hellais:
I am closing this PR as it's now out of sync with master and blocklists are not to be part of the country lists.

PR #35, opened 2016-03-30
Add 3 more major FEXP blocked sites for Malaysia, Add re import for add script.

kaerumy:
add script fails if re module not imported


hellais:
LGTM. Merging.

PR #43, opened 2016-05-05
Clean up malaysian site list my.csv

sweemeng:
* site are remove based on whether domain expired or not
* whether a domain return an answer
* where the site content still exist


sneft:
I believe this is the moved HRW URL:
https://www.hrw.org/legacy/english/docs/2005/01/13/malays9822.htm

hellais:
LGTM.

Issue #27, opened 2015-07-14
Review the proposed new content category codes


sneft:
I have drafted a proposed set of revised content category codes which would replace the existing codes. The proposed set of codes are here:

https://github.com/citizenlab/test-lists/blob/master/lists/00-proposed-category_codes.csv

The aim here has been to streamline the categories, collapsing redundant categories and revising categories which were frequently confusing for list-creators. The goal was simplicity - fewer categories will be easier to use, prevent duplication and ideally limit miscategorization.

Alternative approaches considered were to adopt an existing category scheme wholesale (such as Netsweeper or Blue Coat). These generally have many more categories than the 30 proposed here, which can be a pro or a con depending on your needs.

Please feel free to offer feedback here. This list format is very much building on the prior ONI scheme, so I would love to hear comments based on alternative uses for the lists.

hellais:
It was a pitty to not manage to meet up while in TO!

I noted some of the categories I found missing when importing the BOFH list here (I think this was before the new categories, hence I now realise a lot of them have a new valid category):
https://github.com/citizenlab/test-lists/blob/master/lib/lists/official/it/bofh.py#L46

I will comment in this ticket the ones that I think could still be integrated:

"Portale di link": "MISC"

Which would be something like "Directory", "Link index" (think dmoz, alexa, etc.)

"Phishing": "HACK",

This is a common occurrence in Italy for example where the postal police will block known fishing URLs by ISPs.

 "Riviste": "FEXP", 

This is "Magazines", this could fit within the news category, but it's description should be extended to include also non-news media. Perhaps NEWS could become MEDIA?

"Sito personale": "MISC", #XXX NEWCAT?

This would be "Personal website", like the personal work portfolio or a website meant to be accessed a reduced audience.

sneft:
Yes, I wish we could have discussed this while you were here!

Here's my thoughts on the categories:

Directory, Link Index - I can imagine a category that includes these as well as 'web portals' could be useful. Such portals are particularly popular in Asia. Do you have a specific case in mind when you think of testing such link indexes?

Phishing - One of our general rules has been to avoid intentionally including sites which contain malicious content, so I don't think we'd want to include this.

Magazines - This is a good suggestion. I think we should include this as part of the NEWS category. (MEDIA was my initial choice for a code but I preferred to keep as many to four letters as possible, so it was strictly an aesthetic decision. I'm content either way.)

[...]

PR #44, opened 2016-05-12
Added URLs provided by the Berkman Center

rpanah:
These URLs have been proposed by the Berkman Center for Internet & Society at Harvard. They want to test these URLs for their monthly reports. I've added the categories and made their source "berkman".


sneft:
Hi Abbas,

Thanks for submitting. I see this one is using some of our new category codes (e.g. NEWS) that haven't quite been finalized yet. (See Issue #27) I don't know if this will cause any headaches..

rpanah:
Hi Adam,

I used the proposed categories because they were a better fit. If at some point in the future we end up not using some of these categories, we can update our results and map them to other categories retroactively.

PR #45, opened 2016-05-17
Remove some URLs that are dead

hellais:
These are sites for which the domain has expired and they just point to
parking domain.


jakubd:
Looks good thanks Arturo!

PR #49, opened 2016-06-03
Purging NX domains from: global.csv, ir.csv, tr.csv, cu.csv, sa.csv, kz,csv and vn.csv

chokepoint-project:


hellais:
@chokepoint-project thanks for this! Can you perhaps squash these commits?

anadahz:
@chokepoint-project good work!
If you did this in an automated way it will be nice to share it so that we can do this regularly.

arglebarf:
Its based on the output a purge script which looks at local and remote results per country (country+global) over several weeks of daily checks (dnshonest). So there's not much to share given that its useless without running the entire dnshonest testing tool on a daily basis, which is relatively expensive to do and we're already doing it. I am thinking about hooking it up to a feedback script to push the results directly to git though. So if we're testing that particular country, the cl test-lists would receive purges automatically.

PR #50, opened 2016-06-03
Chokepoint Project commits

chokepoint-project:
Not really squashed, but you should be able to pick the last commit I suppose.


hellais:
We had a discussion about the topic of removing "dead URLs" with @sneft and @agrabeli and came to the conclusion that it's probably ideal to keep testing them even if they are no longer reachable from a certain control vantage point.

The reason for doing so is that while it may be the case that the domain is no longer reachable from a normal vantage point, it may still be blocked in a particular country and hence DNS queries for it will be spoofed and HTTP responses delivered when they trigger the blocked domain in question.

For this reason I think it's best to keep even dead domains inside of the country specific lists, but maybe remove dead ones from the global lists.

@chokepoint-project what do you think?

arglebarf:
The reason we're purging is because running daily checks on domains that don't exists only generates garbage results which have to be filtered out and its already relatively expensive to run daily checks on large amounts of nameserver+domain combinations in several countries, which is what we use the test lists for. (we have nearly 222 million unique query results so far from only 14 countries.)

I can take a look in our data to see if this has actually ever happened during the past three years, but I don't think it has.  While I agree with you in principle, I have the impression that in practice this is a non-issue.

I suggest adding a column to both the global and the country csv files to indicate that the domain has expired. That allows anyone who wants to check those to do so, but does not bother anyone who does not. Also its good provenance maintenance of the test lists data itself.

chokepoint-project:
Will these changes be committed ?

arglebarf:
Just so you know, the url's we propose to remove are really gone. Most have been for a long time. Furthermore we don't purge until the local check has failed for quite some time and for multiple reasons. (such as both the remote DNS servers and the local DNS servers return NX domain). I strongly recommend adding a column 'retired', to be filled with the date the domain is retired. That way those users  who still want to use these domain for testing can do so, and those who do not, do not have not maintain a separate list.

hellais:
I am closing this due to inactivity. If you would still like to contribute them please file a new PR.

PR #72, opened 2016-08-04
Remove asiabynight.com for .my

kaerumy:
Actual content is missing from this site, except for two entries and the comments are mostly spam as is their twitter feed.


kaerumy:
Remove asiabynight.com for .my the site seems to be empty now

PR #79, opened 2016-08-11
Updated domains for popular global torrent sites

kaerumy:


hellais:
This looks good to me!

Thanks!

PR #80, opened 2016-08-27
Add URLs blocked by court order in Norway

graphiclunarkid:
The Oslo District Court has made two orders requiring named ISPs to
block access to lists of p2p filesharing, media sharing and media
streaming websites. This commit includes:
- A .csv in the "official list" format containing all the URLs named in
  both court orders.
- Copies of the court orders (in Norwegian) in .pdf format. The file
  names are case numbers, which are also used in the "source" column of
  the .csv file, so that it's clear which order blocks which URL.


graphiclunarkid:
As discussed in the ooni weekly meeting today, I've pushed another commit creating a Norway-specific URL-list to match the official list, so that ooniprobes in Norway will test these URLs routinely. As far as I can tell from a manual check, the new list is de-duplicated against global.csv

lists/no.csv
```diff
@@ -0,0 +1,75 @@
...
+http://thepiratebay.gd,P2P,File-sharing,2016-08-27,,
+http://thepiratebay.la,P2P,File-sharing,2016-08-27,,
+http://thepiratebay.mn,P2P,File-sharing,2016-08-27,,
+http://thepiratebay.ms,P2P,File-sharing,2016-08-27,,
+http://thepiratebay.mu,P2P,File-sharing,2016-08-27,,
+http://thepiratebay.net,P2P,File-sharing,2016-08-27,,
+http://thepiratebay.se,P2P,File-sharing,2016-08-27,,
+http://thepiratebay.sh,P2P,File-sharing,2016-08-27,,
+http://thepiratebay.tw,P2P,File-sharing,2016-08-27,,
+http://thepiratebay.vg,P2P,File-sharing,2016-08-27,,
```
hellais: Can you fill out the column source as well, with your name of an identifier for the who provided these results?

lists/no.csv
```diff
@@ -0,0 +1,75 @@
+url,category_code,category_description,date_added,source,notes
+http://couchtuner.ag,MMED,Media sharing,2016-08-27,,
```
hellais: Can we just include one variation for the TLD for each site?

Issue #88, opened 2016-10-05
Missing comma's in ae.csv


arglebarf:
The last two entries in ae.csv are missing the final comma for the 'notes' column. Please correct these.

agrabeli:
@arglebarf thanks for pointing this out! I've fixed this: https://github.com/citizenlab/test-lists/pull/90.

arglebarf:
so far so good.

PR #90, opened 2016-10-06
Minor edit based on git issue #88

agrabeli:
Added comas in the last 2 rows, between "sources" and "notes" - as suggested here: https://github.com/citizenlab/test-lists/issues/88.


hellais:
LGTM

PR #94, opened 2016-10-14
Add specific tests for France

Te-k:
Hi,
Here are some first domains to be tested in France.


sneft:
Can you remove the excess spaces in the notes field for some of these URLs?

Te-k:
Is it better now?

PR #99, opened 2016-11-07
New Gambia test list

agrabeli:


hellais:
These are missing the `http://` prefix.

agrabeli:
@hellais I've added the http(s) prefixes. I think gm.csv is ready to be merged.

hellais:
LGTM

PR #114, opened 2017-01-19
Updated EG lists with Signal URLs

sneft:


lists/eg.csv
```diff
@@ -153,3 +153,5 @@ https://twitter.com/ikhwanweb,FEXP,Free expression and media freedom,2014-04-15,
 https://twitter.com/muhammadmorsi,FEXP,Free expression and media freedom,2014-04-15,citizenlab,
 https://www.facebook.com/ikhwanweb.official,FEXP,Free expression and media freedom,2014-04-15,citizenlab,
 https://www.alaraby.co.uk,FEXP,Free expression and media freedom,2016-10-23,OONI,
+https://whispersystems.org,COMT,Communication Tools,2017-01-19,citizenlab,
+http://textsecure-service-ca.whispersystems.org,COMT,Communication Tools,2017-01-19,citizenlab,Used by Signal desktop,
```
hellais: This service appears to not be listening on port 80, but only on 443 (you should use https instead of HTTP)

jdcc: Row 157 has an extra comma at the end, which is breaking some parsers (GitHub's included).

sneft: I'm actually adding this URL to confirm middlebox blocking - I've anecdotally seen the http version blocked by not https, so I'm not actually expecting a legit response from that server. Adding both to be sure. 

Fixed comma, thanks @jdcc

hellais: @sneft sorry only now read the comment here. I see your point. You mean that some HTTP proxy is intercepting the HTTP request and returning some blockpage on that? That would be interesting and I guess it's something worthwhile to test for.

PR #115, opened 2017-01-19
Added EG URL and fixed comma

sneft:


lists/eg.csv
```diff
@@ -155,3 +155,4 @@ https://www.facebook.com/ikhwanweb.official,FEXP,Free expression and media freed
 https://www.alaraby.co.uk,FEXP,Free expression and media freedom,2016-10-23,OONI,
 https://whispersystems.org,COMT,Communication Tools,2017-01-19,citizenlab,
 http://textsecure-service-ca.whispersystems.org,COMT,Communication Tools,2017-01-19,citizenlab,Used by Signal desktop,
```
hellais: I would also remove this line since we would be testing something that is not reachable.

lists/eg.csv
```diff
@@ -155,3 +155,4 @@ https://www.facebook.com/ikhwanweb.official,FEXP,Free expression and media freed
 https://www.alaraby.co.uk,FEXP,Free expression and media freedom,2016-10-23,OONI,
 https://whispersystems.org,COMT,Communication Tools,2017-01-19,citizenlab,
 http://textsecure-service-ca.whispersystems.org,COMT,Communication Tools,2017-01-19,citizenlab,Used by Signal desktop,
+https://textsecure-service-ca.whispersystems.org,COMT,Communication Tools,2017-01-19,citizenlab,Used by Signal desktop
```
hellais: I would also maybe add something in the note mentioning that the cert is not signed by any browser accepted ca, but is signed by a whispersystems ca.

hellais:
This looks good. It needs to be rebased onto master though. I can do it while merging it.

hellais:
Manually applied this diff in master

PR #123, opened 2017-02-06
New URLs for Thailand.

kay-wong:
*Add government websites that may block foreign IP address — see “May block foreign IP addess” in notes.

*Updated with "Meitu Makeover” album from Shanghaiist


hellais:
Thanks for your contribution! 🎉

PR #126, opened 2017-02-09
removing state.gov websites from the list

mrphs:
state.gov is not neutrally considered a human rights website globally and in certain places of the world might buy the tester some extra ordinary and unnecessary troubles that we could simply avoid.

I suggest removing them from the list.


sneft:
Yes, the categorization of the State Department's site as HUMR (Human rights) is clearly incorrect. In general we might argue that the specific website of the DRL does constitute Human rights content, as that is the theme and content contained at that site.

However, with the introduction of the new category codes (https://github.com/citizenlab/test-lists/blob/master/lists/00-LEGEND-new_category_codes.csv) we have a new dedicated content category which may be more appropriate:

GOVT - Government - Government-run websites, including military sites.

Can you rebase and submit again?

Thanks!
Adam

hellais:
I am closing this due to inactivity. If you would still like to contribute them please file a new PR.

PR #127, opened 2017-02-09
removing some items

mrphs:
removing some website that don't exist anymore

### DIFF
--- a/lists/ir.csv
+++ b/lists/ir.csv
@@ -1,7 +1,4 @@
 url,category_code,category_description,date_added,source,notes
-http://213.4.130.210,FEXP,Free expression and media freedom,2014-04-15,citizenlab,
-http://aadaamaak.blogfa.com,GAYL,Gay-lesbian,2014-04-15,citizenlab,
-http://ablogformen2.blogspot.com,GAYL,Gay-lesbian,2014-04-15,citizenlab,
 http://abujahl.blogspot.com,REL,"Religious conversion, commentary and criticism",2014-04-15,citizenlab,
 http://acetaminophen.persianblog.com,POLT,Political transformation,2014-04-15,citizenlab,
 http://alfie-atkins.blogspot.com,POLT,Political transformation,2014-04-15,citizenlab,


sneft:
Can you rebase with the current master?

jakubd:
thanks!  we went and refactored the structure after your contribution though, I have applied these fixes in #136

mrphs:
Hey, sorry for my lack of responsiveness on this ticket... For future reference, it might make sense to have a little script running somewhere to check these links on a regular basis to make sure they're not broken. I've only checked a few handful of them as I was playing around.

jakubd:
No problem, I think this is quite a good idea, all lists like this will eventually go stale.  I am thinking maybe something as a simple rolling visit to the sites across all lists which will flag sites that consistently give errors out would be ideal for this use-case.  Sites can go down and should still stay on the list but if a site goes down over many days then it is likely no longer suitable here. I will try to work on this in the coming days and share.

sneft:
One of the complicating factors is that there are URLs which would be flagged as giving errors (404s, unregistered domains) which nonetheless remain blocked in a given jurisdiction. (Much less consideration is likely given for taking sites off block lists compared to putting them on). One of the rules-of-thumb I've used is that if I know a URL is blocked I will keep it on a list, regardless of the response (or lack thereof) from that URL at the time of testing.

I appreciate that identifying and tracking a dead-and-irrelevant URL versus a blocked-but-404'ing URL is difficult, but I think this is something we should keep in mind. Also, we want to ensure that efforts are made to try to identify content that has been moved rather than just scanning-and-deleting through an automated process.

arglebarf:
For your information, we already automatically check and prune the test lists (and quite conservatively) as discussed under https://github.com/citizenlab/test-lists/pull/50. 

This shows up in our graphs as well https://beta.chokepointproject.net/country/CN#2017-02-01%2023:00

Our database gives this as 'pruning' reasons for the 5 domains mentioned in the graph:

While the given reason is local_nxdomain, the domains would not have become pruning candidates if any of the remote nameservers would have given a different response from the local domain server nor when an http request would have determined that something other than DNS issues might be occurring.

Additionally we need a minimum amount of reports before pruning proceeds and the pruning condition needs to persist for a least 3 days. 

# date_end, date_end_reason, fqdn, country_code
'2017-01-31 23:31:24', 'automatically pruned: local_nxdomain', 'delicious.com', 'CN'
'2017-01-31 23:31:24', 'automatically pruned: local_nxdomain', 'www.bbcworld.com', 'CN'
'2017-01-31 23:31:24', 'automatically pruned: local_nxdomain', 'www.famouspornstars.com', 'CN'
'2017-01-31 23:31:24', 'automatically pruned: local_nxdomain', 'www.sexy-geschichten.at', 'CN'
'2017-01-31 23:31:24', 'automatically pruned: local_nxdomain', 'www.watchtower.org', 'CN'

Given the response on that merge request I concluded that there was no interest in cleanup of the lists. If I am mistaken about this lack of interest in addressing this issue, please let me know. I would prefer to work on finding a solution integral to citizen lab's test-list and of use to others, rather than doing so outside of the test list. 

For instance we could perform the pruning even more conservatively if doing so could be helpful to others.

hellais:
I am of the same opinion as @sneft, I believe that we need to carefully ensure that we are not dropping sites that are dead, but still being blocked.

I think that the methodology used by @arglebarf is still not sufficient as, correct me if I am wrong, it's only based on observations based on observations on open DNS resolvers and doesn't really capture the full picture of blocking.

As an example, just casually looking at the the changes proposed in #50 I found numerous sites that you flag as to be removed, because your DNS based observations show no censorship, but measurements from the country actually show the site to be blocked.

For example:

* https://github.com/citizenlab/test-lists/pull/50/files#diff-67f77d9a16fe08f250586ca8bf334b15L303 is actually blocked in IR: https://explorer.ooni.torproject.org/website/www.montazeri.ws

* https://github.com/citizenlab/test-lists/pull/50/files#diff-67f77d9a16fe08f250586ca8bf334b15L383 is blocked in IR: https://explorer.ooni.torproject.org/website/www.ufin.org

If you have a list of sites that you believe could be candidates for pruning I can maybe cross reference it with OONI data and see if in fact the site in question actually safe to remove it.

arglebarf:
Hm, www.montazeri.ws is no longer locally giving nxdomain, but has been grabbed by sedoparking. I see how a blocking list is also likely to become outdated and continue to block non-existent sites. What that does to the relevance of determining their continued blocking might be a different discussion. We are no longer interested at that point, but that is also a resource issue to some degree. 

Our observations do a little more than looking at DNS responses only, but certainly do not cover a full picture of blocking. However the pruning does not occur based on 'seeing no censorship', but based on the site no longer existing for at least 3 days provided there are enough reports during that period. Had the sedoparking occurred quicker www.montazeri.ws would not have been pruned (falsely not been pruned :) ).  This is something we could make more forgiving by for instance pruning after 7 days, or 30 days. Whatever folks feel most comfortable with.

As I suggested previously, I would prefer to add a field to the csv files flagging the domains with a non 0 value when it no longer exists or is no longer serving the content they were added to the list for.  

Alternatively I can also stop proposing removals here and provide a link so one of you could periodically see what we have pruned and for what reason.

I understand that "http://10.10.34.34?type=Invalid" indicates that there is a blocking page present. To me this suggests that the blocker's blocking list is outdated and that the OONI response can be interpreted as a false positive, but again, that more a question of what one wants to look for. To identify 'censorship' it seems there does not seem much point, to identify blocking equipment's behavior, more point.

How to interpret the OONI probe data for these examples is another issue which I'd be interested in discussing, but this might not be the right place for that. @hellais, you have my contact info, if you have a moment and interest to discuss this, let me know.  

Attached a full list or domains we have currently determined to be dead.
[pruned_domains.csv.zip](https://github.com/citizenlab/test-lists/files/800272/pruned_domains.csv.zip)

sneft:
Thanks for sharing this list of pruned domains - it's very helpful.

I'd also like to add that one other consideration which is that in many cases the content of a given URL may have simply moved elsewhere. For example, in the list of dead domains above, several of these have simply moved (e.g. delicious.com to del.icio.us; www.watchtower.org to https://www.jw.org/en/publications/magazines/). I think this automated scan is very helpful for identifying such domains/URLs, but clearly there's a benefit to replacing any pruned domain with the new domain to which it has moved.

Perhaps we can use your script @arglebarf as a starting point to clean the lists up. This may take some time, and require local/regional expertise, but I think this would be a good first start. My colleague @jakubd also had a suggestion about how we could incorporate these automated scans moving forward. Jakub?

Also - I think it would be a reasonable guideline that the global list should not contain any dead domains. If we find a dead domain on the global list which is blocked in a given country, it makes sense to prune it from the global list and move it to that country list. I believe @hellais mentioned this elsewhere, and I support the idea.

jakubd:
Hey all,

Interesting discussion, a lot of cases are being brought up that I had not considered.  Thanks to everyone for taking the time to write them out.

I think we clearly need to balance our need to retain technically-down-but-still-useful entries and our need to keep the testing list as fresh, lean, and relevant as possible.  Every new URL on this list represents a percentage of someones time when doing a test, and that time is valuable so having the lists be lean is a goal we should not lose sight of as well.  One possible way we could strike this balance is to have a subdir on the repo which named testlist-quality where we could store input such as pruned-domains.csv file that @arglebarf shared or similar tests. Basically we could store any data we have from rolling tests regarding the quality of the list as we have it.  These could serve as an indicator of which URLs we should investigate and focus our resources on investigating.  So when we, or any volunteer who wishes to contribute, wants to see where the most work is needed they have a short list of URLs they can start from.  

For example, if we see that http://www.watchtower.org/ is suddenly down we can do a quick check and see that it got wrapped into http://JW.org and then make a determination whether to remove it or not on that basis.  I think adding a column to each URL is not strictly necessary because we want to minimize schema changes when possible but also it is not related to the actual testing so much as the curation of the list.  Having a subdir will also give us additional flexibility to add and remove tests over time. I think the fine tuned details of how long is a correct threshold to be considered "dead" will depend on the use case but we are obviously far better for having that data than not.

I think a case could also be made for URLs that have been documented as being blocked to stay on the list for a far greater period afterward (months if not longer) for the reasons @hellais and others have said.  This is just an idea, but am interested to hear other perspectives.

-jakub

arglebarf:
Thanks for the responses, I'm relieved to see there is interest in keeping the lists clean.

I think the focus should be on keeping the lists a usable as possible for the various test requirements. I'm a little apprehensive about using a secondary set of 'quality lists' nested in a folder (if I understand the suggestion correctly). It would mean having to do additional parsing and merging of the corrections that it would contain for those who want to use the 'quality lists', as well as the additional work on maintaining those complementary lists. I'm afraid it will be a bit of a headache.

While I agree that schema changes should be avoided, we are talking about csv files here. Any parser of these files will either use the column number or the column name as derived from the header to parse data. As long as any columns are added after the last current column there should not be any problem in either scenario. With such an added column I could automate pushing the pruning output from our systems to present merge requests in the format of Citizen Lab's test-lists, assuming such as thing would be appreciated. (we currently have an 'end_date' and an 'end_reason', when the former is not null our systems ignore the entry for further testing.)

Additionally anyone, such as @hellais, can simply ignore this change. However those who have test cases where expending resources on domains which no longer exist is wasteful, can add a selection based on this column. It would also provide a builtin history of what has been tested in which country over time and what the attrition rate of these domains might be, making the test-lists more useful for longitudinal studies.

I think this test-list is very valuable by virtue of it being curated by Citizen Lab and used as test input for multiple analytical methods, making it easier to compare results. If the usefulness of the lists degrades for a plurality of methodologies I believe this value would be greatly diminished.

hellais:
https://github.com/citizenlab/test-lists/pull/127

PR #132, opened 2017-02-10
added proper Signal server URLs to .eg list

jakubd:
Added the correct URLs for Signal Desktop testing please review to make sure specifying the port doesn't break anything you have before merge.  Signal desktop does TLS over port 80, 4433, and 8443 so we test all these as well as the server used to send attachments.

Also this might make more sense to move to the global list in the future but I think we should clean up that list before adding to it.


hellais:
This should not be a problem for ooniprobe or measurement-kit tests.

I will let other censorship platforms some time to comment before merging.

rpanah:
@jakubd ICLab handles this just fine.

sneft:
Since this works for OONI and ICLab, I'm going to merge.

PR #133, opened 2017-02-14
Updated/Reviewed the Global test list

agrabeli:
I've reviewed all of the URLs included in the global test list and have made the following main changes:

1. Deleted pornographic websites that aren't highly ranked on Alexa.
2. Deleted websites that are squatted/removed/no longer exist/dead.
3. Replaced old domains with their current versions.
4. Added HTTPS to sites that now support HTTPS and load with it automatically when the user accesses them from a normal browser.


jakubd:
Nice!

PR #134, opened 2017-02-15
Update Global with UD

betoantonioplus:
Ecuadorian NGO site Usuarios Digitales that fight for freedom of expresion, digital and human rights and internet gobernance and cybersecurity


hellais:
Are there some particular regions or countries that it would be most relevant to include this site inside of them instead of putting it in the global list?

We have generally been adding to the global list only very high profile sites (that are for the most part of the alexa top 1 million), while it's much easier to merge them into the country specific lists.

hellais:
Thanks for the PR! I am closing due to inactivity. If this is still relevant please open a new PR.

PR #135, opened 2017-02-17
Added URLs to global test list

agrabeli:
I've added the following URLs to the global test list:

1. https://www.frontlinedefenders.org/
2. https://tacticaltech.org/
3. https://bouncer.ooni.io (to see if OONI's bouncer is blocked)


jakubd:
Looks good

PR #140, opened 2017-03-09
Added more urls to BI list

Tafiti:
This list draws largely from alexa listing top 500 and a few from Burundians.

PR #141, opened 2017-03-13
Improve the list linting script

hellais:
Prior to this change, only one of all the exceptions would be printed when linting the list.

Now it will gather all the errors and print them all out.


hellais:
I am going to merge this so that we can end the ping pong in PR #140

PR #142, opened 2017-03-19
Add Cyprus test list

anadahz:


hellais:
Two things:

1. I think that having a testing list that is just 1 category and is just pulled in from the official country list will introduce a bias in the testing.
Dumping official blocklists is great and interesting, but they should be archived inside of the `offical/<country_code>` tree and not be the main testing list of the country.
Ideally we would have inside of this list also sites that cover socio-political topics relevant for the country in question.

2. There are many URLs in the list that are equivalent. I understand that they are not exactly the same URL, but maybe it's best to select a representative subset of them. Keep in mind that every URL added is an amount of time that a user invests in checking the accessibility of it and having redundancy means that their effort is duplicated. See for more context these discussions: https://github.com/citizenlab/test-lists/pull/80, https://github.com/citizenlab/test-lists/pull/127#issuecomment-282398648.

The CI tests are also failing. I thinks it's due to some formatting issues with the list.

anadahz:
Thanks for reviewing the test list @hellais.

My reply to your comments inline below:

> I think that having a testing list that is just 1 category and is just pulled in from the official country list will introduce a bias in the testing.
Dumping official blocklists is great and interesting, but they should be archived inside of the offical/<country_code> tree and not be the main testing list of the country.

That's a great feature but at the moment it's unused.

> Ideally we would have inside of this list also sites that cover socio-political topics relevant for the country in question.

That will be really nice and I hope this can attract some attention, however every test list has started from what was known to be blocked. In the case of Cyprus the people at the Cyprus Observatory determined that this test list will be a representative sample of the [official blocking list](http://blocking.nba.com.cy/) that (at the time of this writing) includes more that 2500 URL entries.

> There are many URLs in the list that are equivalent. I understand that they are not exactly the same URL, but maybe it's best to select a representative subset of them. Keep in mind that every URL added is an amount of time that a user invests in checking the accessibility of it and having redundancy means that their effort is duplicated. See for more context these discussions: #80, #127 (comment).

I understand that every URL adds an extra weight as a user from this country will perform a test on the specific entries, however the current URL list can help understand the following things:

- Identify the blocking infrastructure and how it's being implemented.
- Detect cases of overblocking and underblocking as the list contains URL entries that are not anymore listed in the official blocking list.
- Detect cases of DPI, or misconfiguration on the filtering equipment (404 errors, proxy responses)

The following links provide detailed documentation on the importance of testing similar URL entries.

* https://ooni.torproject.org/post/eeep-greek-censorship/
* https://www.usenix.org/system/files/conference/foci15/foci15-paper-ververis-updated-2.pdf

The CI is happy again there was a URL with a comma that needed to be encoded properly.

agrabeli:
I agree with @hellais that it would be even more useful if this test list includes some diversity in terms of the types of URLs included. Currently, it includes only gambling sites... While it's useful to test gambling sites (especially since it's known that they're blocked in Cyprus), it's also useful to test a wide range of other types of sites to (a) limit bias and (b) conduct a more comprehensive study.

Please add more sites that fall under the following (31) categories: https://github.com/citizenlab/test-lists/blob/master/lists/00-LEGEND-new_category_codes.csv

hellais:
I am closing this due to inactivity. Feel free to reopen another pull request if needed.

PR #149, opened 2017-04-10
Added site to MM test list

agrabeli:
Just added a site (http://www.bestbuy.com/) that was reported to be inaccessible in Myanmar.


hellais:
LGTM

PR #150, opened 2017-04-11
Added sites to VE test list

agrabeli:
I've updated the Venezuelan test list to include the following news sites that have been reported to be blocked:

* http://elcapitolio.tv/
* http://vpitv.com/
* http://cnnespanol.cnn.com/


hellais:
LGTM

PR #152, opened 2017-04-11
Add Argentina test list

lucciferri:


hellais:
Thanks for contributing this test list.

The format however has some issues:

1. The columns are tab separated instead of being comma separated
2. You are using the windows/DOS carriage return instead of newlines to separate the rows

What editor did you use to create this? It's probably best if you use one of the other CSVs as a reference to create the new country test list file.

lucciferri:
Hi Arturo!

Thanks for pointing that out. I've fixed the CSV and updated the date format to match the other lists.

lists/ar.csv
```diff
@@ -0,0 +1,200 @@
...
+http://madres.org,MISC,Miscelaneous content,2017-07-04,ADC
+http://www.riojalibre.com.ar,NEWS,News Media,2017-07-04,ADC
+http://www.laizquierdadiario.com,NEWS,News Media,2017-07-04,ADC
+http://partidopirata.com.ar,POLR,Political Criticism,2017-07-04,ADC
+http://www.mapuche.info,NEWS,News Media,2017-07-04,ADC
+http://www.elorejiverde.com,NEWS,News Media,2017-07-04,ADC
+http://www.jachallaopinion.com.ar,NEWS,News Media,2017-07-04,ADC
+http://www.jachalmagazine.com.ar,NEWS,News Media,2017-07-04,ADC
+https://vimeo.com/205117580,MISC,Miscelaneous content,2017-07-04,ADC
+http://www.cha.org.ar,MISC,Miscelaneous content,2017-07-04,ADC
```
hellais: It would be best to limit the usage of the MISC category code. As a matter of fact the MISC category code was added just to be able to support mapping legacy codes to the new scheme, it's usage for new test lists is highly discouraged.

hellais:
Ok, we almost there.

Still todo:

* [ ] Add a proper category code that is not MISC to the URLs labelled as such

* [ ] Add an extra column to the .csv files (the notes column)

lists/ar.csv
```diff
@@ -0,0 +1,200 @@
+url,category_code,category_description,date_added,source,notes
+http://lanacion.com.ar,NEWS,News Media,2017-07-04,ADC
```
hellais: An extra comma is needed on every line below

lists/ar.csv
lucciferri: What would be the best category for human rights NGOs?

hellais: `Human Rights Issues	HUMR`
https://github.com/citizenlab/test-lists/blob/master/lists/00-LEGEND-new_category_codes.csv#L7

lucciferri:
The latest commit fixes MISC categories and adds the extra column (notes).

hellais:
Almost there:

The CI is complaining about the following:
```
lists/ar.csv (line 24): Duplicate URL "http://www.cronista.com"
lists/ar.csv (line 56): Invalid Category Description "News Medi"
lists/ar.csv (line 100): Invalid Category Code "NEW"
lists/ar.csv (line 121): Duplicate URL "http://www.perfil.com"
```

Also it seems like something went bad in merging master. Don't worry too much about that I can always cherry pick you commit and merge it.

hellais:
Looks good now. I will proceed with fixing up the git history and merging.

lucciferri:
Thank you, @hellais !

hellais:
Merged into master. Thank you so much for your contribution!

PR #153, opened 2017-04-12
Add Italian URLs list

ORARiccardo:
I would like to add a list of websites blocked by the Italian Authority on communications (AGCOM¹) which is blocking them due to copyright infringement. Usually they provide a block page for ISPs which you can find here². Unfortunately Italian ISPs seldom use this block page. 

In addition I provide some URLs of the major Italian news media companies.

¹ https://www.agcom.it/
² https://www.agcom.it/documents/10179/7141049/Allegato+23-3-2017+1490254512337/8fad9ae3-c3e1-416f-85e1-e46dc6b44199?version=1.0


lists/it.csv
```diff
@@ -0,0 +1,333 @@
...
+http://www.agi.it,NEWS,News Media,2017-04-12,,
+http://www.ansa.it,NEWS,News Media,2017-04-12,,
+http://www.quotidiano.net,NEWS,News Media,2017-04-12,,
+http://www.milanofinanza.it,NEWS,News Media,2017-04-12,,
+http://www.leggo.it,NEWS,News Media,2017-04-12,,
+http://city.corriere.it,NEWS,News Media,2017-04-12,,
+http://www.dnews.eu,NEWS,News Media,2017-04-12,,
+http://www.tgcom24.mediaset.it,NEWS,News Media,2017-04-12,,
+http://www.wired.it,NEWS,News Media,2017-04-12,,
+http://www.vice.com/it,NEWS,News Media,2017-04-12,,
```
hellais: Could you add the source column as well? It can be anything including anonymous if you don't want to be credited or some moniker of your choosing.

hellais:
First, so sorry for getting back to you on this so late.

Secondly, I an a bit concerned about adding mostly a dump of the official italian testing list to the official country test list.

Maybe we can try to make a sampling of some of the sites inside of the official blocklist and add a subset of them to the blocklist.

Something like:

* 30 sites from the FILE category
* 30 sites from the GAMBLING category (I see no gambling sites are added, though that is present in the official italian blocklists from AAMS)
* The existing news websites you added
* Some sampling of politically sensitive sites (NOTAV, m5s, etc.)
* Sites critical to religion
* And maybe something else

lists/it.csv
ORARiccardo: added some sources + my moniker on some others

ORARiccardo:
Totally agree with you, I'll re-sample the websites, definitely missed some GAMBLING. Really important to add politically sensitive websites and religion ones.

erciccione:
Can this PR be merged? We may update it in a second one

hellais:
> Can this PR be merged? We may update the it section with a second PR

Ok I am going to merge this, thought it would be great if we could populate it with some more websites as well from more diverse categories (ex. collective, political parties/movements, human rights topics, communications platforms, etc.)

PR #159, opened 2017-05-15
Add wannacry(pt0r) ransomware domains

anadahz:


hellais:
Is this still relevant?

Are we sure we really want to be testing this in ooniprobe and other tools? I get the impression that it could lead to some bad side-effects for users (being suspected of being infected by malware) and maybe it's not really that worth to risk that for minimal gain (what is the benefit of learning that they are blocking malware domains?).

anadahz:
One variant of the malware first checks for the "kill switch" domain name and if it is not found (eg: DNS "kill switch" domain entry pointing to `127.0.0.1`) then  the ransomware gets [initiated](https://en.wikipedia.org/wiki/WannaCry_ransomware_attack#Cyberattack).

Although submitted long time ago, I think it is still relevant since many DNS servers were (and perhaps still) answer with the record `127.0.0.1` to the ransomware's "kill switch" domain name query.

hellais:
I see, though there are concerns with having probes behaving as if they were infected by malware and possibly raising unwanted suspicions.

anadahz:
Feel free to close this PR if it's too risky for probes.

PR #161, opened 2017-05-18
Updated Iranian test list

agrabeli:
The Iranian test list has been updated thanks to URLs provided from the folks at ASL19. However, the categories probably require some review, since I added them myself (based on Google translations...not being a Farsi speaker). A review of the categories would be much appreciated!


hellais:
LGTM

PR #163, opened 2017-05-25
Added news websites blocked today

agharbeia:
These websites are blocked as of today, only one ISP in Egypt seems to be unaffected by the blockage. The websites are accessible via proxy at the time of this request.


hellais:
LGTM

anadahz:
> These websites are blocked as of today, only one ISP in Egypt seems to be unaffected by the blockage. 

@agharbeia Do you know which ISP seems to not be affected by the blockage?

agharbeia:
It's http://noor.net again
AlMasry AlYoum published a report by an unnamed security agency, i.e intelligence, justifying the action and citing similar cases in other countries, and claiming it as a step against terrorism: http://www.almasryalyoum.com/news/details/1139015
I can access one site through https, while others say they cannot!

anadahz:
@agharbeia thanks for the heads up!

Do you think that the domains included here are relevant (?): https://github.com/citizenlab/test-lists/pull/164

agharbeia:
@anadahz, the domains added in #164 are also relevant. Thanks.

PR #166, opened 2017-05-25
Added 1 url

Tafiti:


sneft:
I notice that this URL 404's. Has it been moved elsewhere? Could you add more context?

Tafiti:
Turns out that url has been brought down by the media company.
This may help us understand trends around self-censorship. It was reporting on state procurement of Internet surveillance equipment.

Should we keep the url on the list?

hellais:
> Turns out that url has been brought down by the media company.

I checked on archive.is and it seems like the URL was not archived. I think it would be a best practice to always archive these URLs via http://archive.is/ when they are added, perhaps we could even do this automatically.

> This may help us understand trends around self-censorship. It was reporting on state procurement of Internet surveillance equipment.
> Should we keep the url on the list?

I am not sure if at this point it makes sense to add it to this particular testing list, since we don't really care to have users check the reachability of it. It maybe interesting, though, to start some other list repository for storing these sorts of URLs (with the connected archive.is snapshotting of the URL).

Tafiti:
I like the idea of using archive.is to capture the moment in time. 

Also, worth checking [this](https://petertodd.org/2017/carbon-dating-the-internet-archive-with-opentimestamps) idea by @petertodd on timestamping archive.is.

PR #169, opened 2017-05-28
somemore over the past two days

agharbeia:
Three news sites (added their work-around aliases too), one human rights site, and [bridges.]torproject.org

### DIFF
--- a/lists/eg.csv
+++ b/lists/eg.csv
@@ -25,6 +25,8 @@ http://bokra-a7la.blogspot.com,NEWS,News Media,2014-04-15,citizenlab,
 http://cairofreeze.blogspot.com,NEWS,News Media,2014-04-15,citizenlab,
 http://www.cairoportal.com,NEWS,News Media,2017-05-24,ADEF,blocked as of date added
 http://changeegypt.blogspot.com,NEWS,News Media,2014-04-15,citizenlab,
+http://dailynewsegypt.com,NEWS,News Media,2017-05-28,ADEF,blocked as of date added
+http://thedailynewsegypt.com,NEWS,News Media,2017-05-28,ADEF,blocked as of date added
 http://digressing.blogspot.com,NEWS,News Media,2014-04-15,citizenlab,
 http://dostor.org,NEWS,News Media,2014-04-15,citizenlab,
 http://egyptiancitizen.blogspot.com,NEWS,News Media,2014-04-15,citizenlab,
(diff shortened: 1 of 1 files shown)


hellais:
LGTM. Thanks!

lists/eg.csv
```diff
@@ -56,15 +60,19 @@ http://misrdigital.blogspirit.com,NEWS,News Media,2014-04-15,citizenlab,
...
 http://samy.attahawy.net,NEWS,News Media,2014-04-15,citizenlab,
 http://seekingfreedom.blogspot.com,NEWS,News Media,2014-04-15,citizenlab,
+http://www.sinaihr.org,HUMR,Human Rights Issues,2017-05-28,ADEF,blocked as of date added
 http://tadwen.com,NEWS,News Media,2014-04-15,citizenlab,
 https://textsecure-service-ca.whispersystems.org:80,COMT,Communication Tools,2017-02-10,citizenlab,Used by Signal Desktop for communication
 https://textsecure-service-ca.whispersystems.org:4433,COMT,Communication Tools,2017-02-10,citizenlab,Used by Signal Desktop for communication
 https://textsecure-service-ca.whispersystems.org:8443,COMT,Communication Tools,2017-02-10,citizenlab,Used by Signal Desktop for communication
 https://whispersystems-textsecure-attachments.s3.amazonaws.com,COMT,Communication Tools,2017-02-10,citizenlab,Used by Signal desktop for attachments
 http://theegyptblog.blogspot.com,NEWS,News Media,2014-04-15,citizenlab,
+https://torproject.org,ANON,Anonymization and circumvention tools,2017-05-27,ADEF,blocked as of date added
```
anadahz: This link is already in the [global list](https://github.com/citizenlab/test-lists/blob/master/lists/global.csv#L981).
In ooniprobe the global list is always tested regardless of the localization.

lists/eg.csv
```diff
@@ -56,15 +60,19 @@ http://misrdigital.blogspirit.com,NEWS,News Media,2014-04-15,citizenlab,
...
+https://bridges.torproject.org,ANON,Anonymization and circumvention tools,2017-05-28,ADEF,blocked as of date added
```
anadahz: I will suggest you to add this entry in the global list instead, as it will be useful to be tested in all countries.

agharbeia: Sure. I thought so after having requested the merge. I hadn't checked the global list back then, bt now I did. "www.torproject.org" is already there. Do you suggest that I add "bridges.torproject.org" there as well?
Once I do I will remove torproject.org from here. I also intend to remove a couple of sites that are no longer published.
Also what's the policy on "www" when its presence and absence work? Sometimes blocking affects one case and not the other.

anadahz: Indeed the entry `bridges.torproject.org` will be useful in the global lists.
Indeed there are different cases being tested with/without the `www` prefix.
@agharbeia thanks for contributing to the test-lists!

PR #171, opened 2017-05-31
Moved tor sites to global, and refreshed others

agharbeia:
Removed torproject.org from eg.csv as it exists in the global list, and moved bridges.torproject.org there as well.
Removed a couple of sites that are no longer published at the listed domain names.
Added a couple of more news sites as candidates for blocking in Egypt to eg.csv


agharbeia:
I'll make another one including later commits.

PR #173, opened 2017-06-01
Moved Tor domains to global, and added and refreshed others

agharbeia:
Removed torproject.org from eg.csv as it exists in the global list;
Moved bridges.torproject.org there as well;
Added OONI related domains
Removed a couple of sites that are no longer published at the listed domain names.
Added a couple of more news sites as candidates for blocking in Egypt to eg.csv


hellais:
LGTM

PR #178, opened 2017-06-12
Updated AZ test list

agrabeli:
Added 2 more sites to the AZ test list that were recently reported to be blocked.


sneft:
Looks like there's a duplicate of the abzas.net URL.

agrabeli:
I have addressed the feedback. Please merge.

PR #180, opened 2017-06-19
Added news site to Palestinian test list

agrabeli:
The news site added to this list was recently reported to be blocked locally.


hellais:
It seems like all the URLs suggested in this PR have already been added in: https://github.com/citizenlab/test-lists/pull/183. I think this can be closed.

PR #181, opened 2017-06-20
Updated Egyptian test list

agrabeli:
I updated the Egyptian test list to include sites that were reported to be blocked (ref: https://afteegypt.org/right_to_know-2/publicationsright_to_know-right_to_know-2/2017/06/04/13069-afteegypt.html?lang=en).


hellais:
Seems like the github rebase did not actually work. I would suggest at this point to open a new PR with just the added URLs in it.

agrabeli:
@hellais I opened another pull request #194 so closing this one.

PR #184, opened 2017-06-23
Created a test list for Mongolia

agrabeli:
This new test list includes Mongolian news websites, portals, human rights sites, LGBT sites, etc. 

I've also included a few sites from Mongolia's official blocklist: http://black-list.mn/?p=2 

Most sites from the blocklist though are no longer operational (e.g. parked domains).


hellais:
I see that most of the sites in this list are actually on the mongolia tld (`.mn`). I think it's generally more useful to test for sites outside of the TLD of the country (if they want to block domains on the TLD of the country they can just do a takedown of the domain).

I would suggest maybe taking more domains from the blacklist and replacing some of the `.mn` tlds.

hellais:
Bump.

agrabeli:
@hellais Most of the URLs included in the official blocklist are no longer operational (parked domains, servers are down, etc etc), I've included the URLs which are still operational and not already included in the global list. It's worth noting that the official blocklist includes many .mn domains, suggesting that they may resort to blocking such domains (as well) - not only (?) take down requests. 

Most of Mongolia's major media websites have .mn domains. 

I'd kindly request that this pull request is merged so that some testing can get started. We can of course continue to update the test list on an ongoing basis.

agrabeli:
@hellais Bumping this up.

agrabeli:
@hellais Consider merging this?

hellais:

PR #190, opened 2017-07-17
Add two URLs to the South Sudan test list

hellais:


hellais:
@jakubd @sneft can you look into merging this?

PR #192, opened 2017-07-19
Created new test list for Mali

agrabeli:
Thanks to URLs provided by one of our partners, we now have a new test list for Mali.


hellais:
LGTM!

PR #199, opened 2017-08-09
adding autistici and noblogs

opensamba:
Hi,
I'm adding this two website to the global list as I think it's worth to know if the A/I services are censored and if so in which country

 * http://autistici.org/
 * http://noblogs.org/

bye


hellais:
Thanks! Merging.

PR #207, opened 2017-08-28
Add 1st part of the CryptoRave activity outcome

lucastx:
After some months in a thread, the voluntaries of the OONI CryptoRave activity have assembled a first batch of URL's to be added to the test list. In short time comes a second one.


hellais:
LGTM.

PR #211, opened 2017-09-05
Updated IR and Global test lists

agrabeli:
As part of this pull request I:
1. Edited some category codes in the Global and IR test lists (some categories were wrong)
2. Updated the IR test list to include URLs (and their respective category codes) that we found to be blocked in Iran based on ooniprobe tests


hellais:
LGTM

PR #213, opened 2017-09-06
Cleaned up the Global and IR test lists

agrabeli:
I replaced "MISC" in the Global and IR test lists with all their respective categories, and I changed category codes that appeared to be wrong (across both lists).


hellais:
Thank you so much for doing the painful job of re-categorising all the mis-categories URLs! 👍

PR #218, opened 2017-09-18
completely revised list

VirtualRoad:
For the upcoming elections in KG please find a new fully updated list. The old list was full of dead sites so a complete merge is not a good idea, we checked every site and deleted all sites that no longer exist and added new ones.


agrabeli:
This PR looks great! 
@hellais please merge it and update ooni-resources.

PR #229, opened 2017-09-26
Added OTF site to Global test list

agrabeli:
It has been reported that the OTF site has been blocked in various countries around the world, and so it's being added to the global test list.


hellais:
LGTM

PR #230, opened 2017-09-27
Fixed category codes in PK test list

agrabeli:
So apparently the problem is that when we converted all FEXP ("Freedom of Expression") category codes to NEWS ("News media"), a bunch of sites that could be categorized as FEXP, but shouldn't be categorized as NEWS, got converted to NEWS.

As a result, we have _many_ sites that shouldn't be categorized as "NEWS" in the PK test list and in many other country-specific test lists (e.g. ir.cvs -- I made changes there in a previous PR). 

I think we can review the categories (and change all the misleading NEWS categories) in country test lists on an ongoing basis...

For now, I have fixed this in the PK test list, which is what this PR includes. 

@hellais Please review and merge this PR asap so that we can generate new information that will feed into the Pakistan report.


hellais:
LGTM

PR #237, opened 2017-10-06
added bbg urls

greatfire-martin:
I suggest adding these BBG content URLs for global testing. There's a total of 220 new URLs which are all different editions of RFA, RFERL and VOA. I think it's important to test the availability of all BBG content URLs globally, not just the local language edition, to fully understand the extent of censorship in each location - ie are they blocking one specific URL, most or all of them. Also, the list includes both HTTP and HTTPS versions of each URL. One may be blocked while the other is accessible.

Since this is a major addition, I have auto-sorted the list according to the sorting logic that seemed to be in place in the current list. This means that a few existing URLs have moved to a new position, since they seemed to not have been sorted in the existing version.

### DIFF
--- a/lists/global.csv
+++ b/lists/global.csv
@@ -12,6 +12,8 @@ http://adium.im,GRP,Social Networking,2014-04-15,citizenlab,Updated by OONI on 2
 http://adultfriendfinder.com,DATE,Online Dating,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 http://advocacy.globalvoicesonline.org,HUMR,Human Rights Issues,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 http://alt.com,PORN,Pornography,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
+http://amharic.voanews.com,NEWS,News Media,2017-10-06,voa,Added by GreatFire.org on 2017-10-06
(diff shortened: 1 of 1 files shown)


hellais:
I am not sure it's going to be optimal to have so many URL from BBG tested globally. I see that some of the sites are localised content, so maybe they are better suited for the respective country lists.

Keep in mind that by adding things to the global testing list, it means that all clients will be testing all of these URLs.

The test-lists are designed to be broad, but not necessarily include everything that is interesting to test for (we need to have some compromises to avoid using up too much of users bandwidth).

I think adding 200+ URLs that are from a single entitity to the global list means that ~25% of the URLs tested are going to be from this group and leads to less diversity in testing.

We would probably be OK with adding a subset of these or integrating those relevant to the specific countries in the country list.

sneft:
I agree with @hellais. Particularly with the dozens of voanews.com and rfa.org URLs - I think these are best included on individual country lists. 

Also, for simplicity it would be best to reduce the value of the 'source' column of the added URLs to simply "Greatfire.org" rather than "Added by GreatFire.org on 2017-10-06".

hellais:
Another thing is also that you don't need to add both `http` and `https` versions of the URL. If the site supports `https` you should specify ONLY the `https` version. OONI Probe and other similar software is already taught to automatically test both, see: https://github.com/citizenlab/test-lists/issues/40

hellais:
I am closing this due to inactivity. Feel free to open another pull request.

PR #238, opened 2017-10-10
Add 2nd part of CryptoRave 2017 activity outcome

semente:
Related to https://github.com/citizenlab/test-lists/pull/207


hellais:
@agrabeli can you take a look at this and us if it's good to merge?

agrabeli:
@semente Thank you for your very comprehensive review of the Brazilian test list, it looks great! :)

I only have 2 questions:

1. Why have many Brazilian governmental sites been added?

2. Two URLs (womenonwaves.org and womenonweb.org) in the Brazilian test list were recently added to the global list (https://github.com/citizenlab/test-lists/pull/248). If we keep those URLs in the Brazilian test list, OONI Probe users in Brazil would be testing those URLs twice each (since the global list is tested by default as well). Perhaps consider removing them from the Brazilian list since they will be tested via the global list anyway?

@hellais This PR looks great and I think it's good to merge (though I would like the above 2 points to be considered first).

semente:
Hi @agrabeli! I will remove womenonwaves.org and womenonweb.org from the br list. About the governmental sites, most are from municipal administrations.

agrabeli:
@semente thanks!

PR #245, opened 2017-10-21
Adds list for the UK

ana-cc:
Initial list, includes some websites meant to be blocked in the UK by court order, as well as websites which are selectively blocked by some ISPs. The URLs were selected via ORG's blocked.org.uk, which tests website blocking against the main UK ISPs.

### DIFF
--- a/lists/uk.csv
+++ b/lists/uk.csv
@@ -1,0 +1,205 @@
+url,category_code,category_description,date_added,source,notes
+http://www.agriyork.co.uk,ALDR,Alcohol & Drugs,19/10/17,,
+http://www.ukcia.org,ALDR,Alcohol & Drugs,19/10/17,,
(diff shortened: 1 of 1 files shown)


sneft:
Some of the category codes need to be corrected. The complete lists of codes (and their descriptions, which must match) can be found here: https://github.com/citizenlab/test-lists/blob/master/lists/00-LEGEND-new_category_codes.csv

As well, the two-letter country code for the UK is actually 'GB', so the file should be called gb.csv. See the country code list here: https://github.com/citizenlab/test-lists/blob/master/lists/00-LEGEND-country_codes.csv

hellais:
BTW the country code for the UK is `GB`.

I think this is a good starting point for a test list for the UK, but I suggest doing it as part of another PR.

I will open a ticket so we don't forget it.

PR #250, opened 2017-10-31
Add abortion & sexual assault sites

willscott:
A few suggested current relevant sites.
Women On Waves and Women On Web are abortion support providers
Rainn is an anti-sexual violence organization


hellais:
This has dupes from: https://github.com/citizenlab/test-lists/pull/248

willscott:
We probably got pointed to them by the same source.  I guess this can serve as a motivator for getting through these pull requests faster to reduce duplication of effort

willscott:
will resubmit a new one with just the rainn addition

PR #252, opened 2017-11-09
Added Protonmail and also Security First Website

rorymbyrne:


hellais:
Thanks for the PR!

As the CI bot says, there is a missing column (you need an extra `,` at the end of each line for the comment column)

hellais:
LGTM. Thanks!

Issue #179, opened 2017-06-15
Archive pages that get added to the test list


hellais:
It would be cool if as part of the CI process we also archived pages via http://archive.is/ as they are added.

This way if the site goes down or is self-censored, we still have an archived copy of it (and we get to know what the site was about when it was still accessible).

archive.is supports something called the memento API that we can use to automate this: http://mementoweb.org/depot/native/archiveis/.

xhdix:
In some Iranian ISPs this address is censored:
![image](https://user-images.githubusercontent.com/12384263/63080980-1eb2d680-bf4b-11e9-8f26-7af20cd74db0.png)
![image](https://user-images.githubusercontent.com/12384263/63081343-0e4f2b80-bf4c-11e9-8b56-adcc96b24389.png)
![image](https://user-images.githubusercontent.com/12384263/63081015-368a5a80-bf4b-11e9-98b8-cc2e9e380756.png)
![image](https://user-images.githubusercontent.com/12384263/63081046-49049400-bf4b-11e9-85a5-619697e7bffe.png)

hellais:
There is now a way back machine public API, here are the docs for it: https://docs.google.com/document/d/1Nsv52MvSjbLb2PCpHlat0gkzw0EvtSgpKHu4mk0MnrA/edit#

PR #267, opened 2017-12-06
Feature/normalise and archive

hellais:
In this branch I add support for what is described in: https://github.com/citizenlab/test-lists/issues/179

For the moment it's not enabled by default as it's very slow to run.

Ideally it would only run the archiving on URLs that haven't been archived recently and I think the best way to do that would be to add columns containing the archived URL (instead of asking the archive.is for it).

I am currently running it against all the citizenlab test-lists repo to archive all the URLs and I expect it to take quite some time to complete.

### DIFF
--- a/scripts/lint-lists.py
+++ b/scripts/lint-lists.py
@@ -57,6 +57,9 @@ class InvalidCategoryCode(TestListErrorWithValue):
 class InvalidCategoryDesc(TestListErrorWithValue):
     name = 'Invalid Category Description'
 
+class NonCanonicalURL(TestListErrorWithValue):
+    name = 'URL is not canonical. It should end with a /'
+
 def get_legacy_description_code(row):
     return row[1], row[0]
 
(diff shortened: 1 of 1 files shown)


scripts/lint-lists.py
```diff
@@ -115,10 +166,11 @@ def main(source='OONI', notes='', legacy=False, fix_duplicates=False):
...
                     )
                 url = url.strip().lower()
-                canonical_url = url
-                if url.endswith('/'):
-                    # We strip trailing / for canonical URLs
-                    canonical_url = url[:-1]
+                canonical_url = get_canonical_url(url)
+                if canonical_url != url and canonical_check:
+                    errors.append(
+                        NonCanonicalURL(url + ' instead of ' + canonical_url, csv_path, idx+2)
```
darkk: It's interesting that you put `csv_path` as separate argument, but format `url` and `canonical_url` into a message string.

hellais: That is done because it's common to all the exceptions. If you look at the structure of the exception handlers, the format is `Exception(human_meaningful_thing, csv_path, csv_file_row)`.

Maybe the cleanest thing would be to create a new exception type for this particular sort of exception, but this seemed good enough.

agrabeli:
@hellais What is the future of this PR? Should we merge it or close it?

mScott224 (approved):

PR #270, opened 2017-12-13
Updated RU list

agrabeli:
I've added a news site (provided by a community member) that's reportedly blocked in Russia.


darkk:
BTW, have you looked at #263 ? It already has openrussia and some URLs from same "website family" like openuni / khodorkovsky.

agrabeli:
@darkk Ah, I hadn't noticed -- thanks for pointing out that PR. Deleting this PR in that case (to avoid duplication).

PR #275, opened 2017-12-17
Update ru.csv

kermoshina:
Added most recent urls reportedly blocked because of the recent "Law on Media as Foreign Agents".


hellais:
LGTM.

PR #277, opened 2018-01-02
Updated global

eddie4:
Added leechers-paradise.org which is an openbittorrent tracker.  Tracker is used for downloading anything and everything. from Linux to probably copyrighted material.

### DIFF
--- a/lists/global.csv
+++ b/lists/global.csv
@@ -744,6 +744,7 @@ http://www.layevangelism.com,REL,Religion,2014-04-15,citizenlab,Updated by OONI 
 https://www.lds.org/,HUMR,Human Rights Issues,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 http://www.learningpartnership.org,HUMR,Human Rights Issues,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 https://www.leaseweb.com/,HOST,Hosting and Blogging Platforms,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
+http://www.leechers-paradise.org,P2P,Peer-to-peer file sharing,2018-01-02,Eddie4,Updated by Eddie4 on 2018-01-02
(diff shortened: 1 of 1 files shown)


eddie4:
Hmm seems i found an old page with wrong catagories. Can't find where i found it tho.

hellais:
I think we already have a fairly representative sample of torrent related websites in the global testing list.

Given the fact that this site is ranked by alexa as number "409,665" I think it's hard to justify including it inside of the global testing list.

That said it seems like it's pretty popular in India, so maybe it could be considered for inclusion in the India test list.

eddie4:
I have no clue how alexa calculates the popularity. But it's not so much the website that is interesting but the service that is running on udp port 6969. This service has 50 million unique visitors a day. It burns through almost 4TB a day of 10kbit of requests. 

It is currently actively banned in Russia and south afrika (That i know of)
The following sites are the backbone of the bittorrent network (yes i know DHT is a thing)
tracker.coppersurfer.tk
tracker.leechers-paradise.org
zer0day.ch
open.demonii.com 
exodus.desync.com

hellais:
> But it's not so much the website that is interesting but the service that is running on udp port 6969.

I see. If that is the case, then adding it to the test list will not yield the desired result.

I can see there being quite some value in having tests also for UDP reachability and in particular for torrent trackers.

As such I filed this ticket in measurement-kit, the library that implement OONI tests: https://github.com/measurement-kit/measurement-kit/issues/1546.

I think we would gladly accept a patch implementing some test of this type and having as one of the tested trackers be `leechers-paradise.org`.

eddie4:
That would be even more awesome. Having actual logic to detect higher level messing with traffic.

hellais:
> That would be even more awesome. Having actual logic to detect higher level messing with traffic.

Excellent. I am going to close this PR as we have the other ticket open about this. Thanks for your interest!

PR #278, opened 2018-01-06
Add fire.news to TW list

pellaeon:
fire.news is the political commentary site by pro-chinese-unification groups in Taiwan, mainly the New Party.

Relevant news of recent arrests of New Party members: http://www.taipeitimes.com/News/taiwan/archives/2017/12/23/2003684482


hellais:
LGTM.

PR #280, opened 2018-01-15
Add anticapitalist.party for testing in turkey, per blocking report

willscott:
https://anticapitalist.party/@accedie/99345033281811972


hellais:
@sneft can you take a look at this PR?

PR #286, opened 2018-01-25
Remove 911lies.org

hellais:
This should be deleted as it's no longer active and we have no evidence of it being blocked based on OONI data in any country.
It was reported by a community member.


hellais:

PR #290, opened 2018-01-29
Updated the KE test list

agrabeli:
I have added URLs provided by community members to the Kenyan test list.


hellais:

sneft:
I think
https://wwwcloudromance.com
should be:
https://www.cloudromance.com

hellais:
@sneft good call! Fixed.

PR #292, opened 2018-01-29
Created new test list for the DRC

agrabeli:
We now have a new test list for the DRC thanks to URLs provided by community members.


hellais:

PR #293, opened 2018-02-01
update Egypt list eg.csv

moeltaher:
added 31 new blocked websites in eg.csv


hellais:
Looks good to me. Merging.

(Side note: we should maybe at some point do a bit of pruning of the `ANON` category sites as I see that the test list for EG contains a disproportionate number of them cc @agrabeli)

moeltaher:
@hellais @agrabeli 
 i tried to reorganize eg.csv. so I placed all the websites related to the "ANON" Category at the bottom of the file.

PR #295, opened 2018-02-03
Reorganize the Egypt list eg.csv

moeltaher:
Reorganize the Egypt list: all the websites related to the "ANON" Category are placed at the bottom of the file.


hellais:
LGTM

PR #296, opened 2018-02-07
Add GitHub to global list

EvasivePangolin:
GitHub hosts many projects that are prime targets for censorship. 
Note: GitHub is blocked in Clarksville-Montgomery County Tennessee (CMCSS) schools, which prompted me to add it to the master list.


hellais:
I edited the URL to be in canonical form (without the www) and replaced the http version with the https version (OONI Probe checks both).

PR #297, opened 2018-02-07
Updated ZW test list

agrabeli:
More URLs provided by community members have been added to the Zimbabwean test list.


hellais:
@sneft could you take a look at merging this PR?

I also added a new label call "review-needed", which basically means these are issues that need a review from somebody other than OONI.

PR #302, opened 2018-02-12
Add pastebin to global test-list

anadahz:


lists/global.csv
```diff
@@ -1114,3 +1114,4 @@ http://zoomshare.com/,HOST,Hosting and Blogging Platforms,2014-04-15,citizenlab,
 https://www.opentech.fund/,HUMR,Human Rights Issues,2017-09-26,OONI,
 https://www.womenonwaves.org/,HUMR,Human Rights Issues,2017-10-26,OONI,
 https://www.womenonweb.org/,HUMR,Human Rights Issues,2017-10-26,OONI,
+http://pastebin.com/,HOST,Hosting and Blogging Platforms,2017-02-12,Anonymous,Often blocked website
```
hellais: Should be added as `https://pastebin.com`

anadahz: Although `http://pastebin.com` redirects to `https://pastebin.com` we may catch more interesting results such as block pages, whereas in the HTTPS version of the website the measurement will most probably time out as block pages are usually not served via HTTPS.

A related OONI repor examplet serving a block page message on the HTTP version of pastebin.com:
https://explorer.ooni.torproject.org/measurement/20150102T020629Z_oqsKJozJnrpVDuLypfyGyecczllhwRlYZRDNmuxDhIapgfcjHh?input=http:%2F%2Fpastebin.com

Thank you for reviewing this pull request.

hellais: See: https://github.com/citizenlab/test-lists/issues/40 for the discussion on why we should include `https` when the site supports it.

anadahz:
Thanks for the mention to #40 changed input entry to HTTPS.

hellais:
(merged master into this and resolved the conflict)

FYI, it's optimal to ping me when you resolve the required feedback, otherwise I will not know that the required fixes have been done.

LGTM, merging.

PR #306, opened 2018-02-26
Add entry to DE test-list

anadahz:


hellais:
This is a great start on the german test list. Maybe we want to wait to make more progress on adding more URLs to the country specific list, before merging this?

anadahz:
@hellais This pull request is open for 19 days, unless you have any new entries to add to the list I will suggest that it should be merged.
I guess nothing stops people from adding entries at a later point.

PR #309, opened 2018-03-02
Add a test list for Lithuania

GiedriusS:
Add a test list for LT (Lithuania). Generated using this hacked together script: https://gist.github.com/GiedriusS/4ba076a7829b8a4c670fa1a63ce77ef1.


hellais:
Hi @GiedriusS,

Thanks for your interest in contributing to the testing lists!

It seems, however, that all the URLs in this list are just gambling websites.

As we explain in the contribution guidelines (https://ooni.torproject.org/get-involved/contribute-test-lists/), that you should maybe check out if you haven't already, testing lists are **not** block lists.

It would be great if you were to contribute to the lithuania testing list and you should include some gambling websites, but also content from different categories.

GiedriusS:
So, if I understand you correctly, I should add some popular sites in Lithuania that might *more likely* to get blocked from other categories, and not just sites which are blocked already? Is that correct?

hellais:
> So, if I understand you correctly, I should add some popular sites in Lithuania that might more likely to get blocked from other categories, and not just sites which are blocked already? Is that correct?

Yes and perhaps drop some of the gambling websites that are known to be blocked and only include a subset of the ones you suggested in this PR.

GiedriusS:
@hellais hi, sorry for taking a long time but I updated the list according to your request and added 2 political sites which might get blocked soon.

GiedriusS:
@sneft ping.

hellais:
Look good to me, thanks for putting this together!

PR #315, opened 2018-03-08
Change GAYL to LGBT in category code list

ecawthon:
GAYL appears only in the category code lists. In all blocklists, this has already
been changed to LGBT, so there is no need to keep this for backwards
compatibility, whereas keeping it in this file is confusing.

This commit also removes the redundant "no pornography" designation, and
replaces it with "community".


hellais:
Hi @ecawthon,

The file you have made changes to is the old category code definition file.

If you look at: https://github.com/citizenlab/test-lists/blob/master/lists/00-LEGEND-new_category_codes.csv#L18, you will see that the category is now called LGBT already.

PR #323, opened 2018-04-01
lists/ir.csv: Add hdl.handle.net

### LINKED ISSUE #322: hdl.handle.net is blocked in Iran, not sure how to categorize
Handle.net is a provider of persistent URLs and is commonly used for institutional repositories (similar to DOIs). I am the administrator of a large open access repository of agricultural outputs and we recently noticed that the Handles do not work in Iran.

Here is an example of a persistent Handle link to an item on our repository: http://hdl.handle.net/10568/91553

I would submit a pull request with the URL in the Iran country list, but I'm not sure what to classify it as.

alanorth:
Handle.net is a provider of persistent identifiers that are used in academic and institutional repositories. Several users have told me hdl.handle.net links using both both HTTP and HTTPS are inaccessible in Iran.

Closes #322.

### DIFF
--- a/lists/ir.csv
+++ b/lists/ir.csv
@@ -969,3 +969,4 @@ http://www.asianangels.com,PORN,Pornography,2017-09-05,OONI,Found to be blocked 
 http://www.linay.com,PROV,Provocative Attire,2017-09-05,OONI,Found to be blocked in Iran based on OONI tests
 http://www.bbcpersian.com,NEWS,News Media,2017-07-19,BBC,
 https://www.radiofarda.com/,NEWS,News Media,2017-09-29,OONI,
+http://hdl.handle.net/10568/91553,SRCH,Search Engines,2018-04-01,Alan Orth,Handle.net provides persistent identifiers for items in academic repositories


lists/ir.csv
```diff
@@ -969,3 +969,4 @@ http://www.asianangels.com,PORN,Pornography,2017-09-05,OONI,Found to be blocked
 http://www.linay.com,PROV,Provocative Attire,2017-09-05,OONI,Found to be blocked in Iran based on OONI tests
 http://www.bbcpersian.com,NEWS,News Media,2017-07-19,BBC,
 https://www.radiofarda.com/,NEWS,News Media,2017-09-29,OONI,
+http://hdl.handle.net/10568/91553,SRCH,Search Engines,2018-04-01,,Handle.net provides persistent identifiers for items in academic repositories
```
hellais: Can you add the author column? You can put `anon` if you don't want to be identified.

alanorth:
@hellais I have force pushed a new commit to my branch with myself as the author. Thanks.

hellais:
lgtm

PR #325, opened 2018-04-06
Correct ill-formed dates in all lists.

zackw:
In several of the lists, there are "date_added" values that aren't parseable as ISO 8601 dates (that is, not in the form YYYY-MM-DD). The most serious such case is a date recorded as "2018-14" with no month, and there were also an entire file's worth of entries in the ambiguous XX/XX/XXXX format.

This patch corrects all of the problem dates and also adds a check for date validity to lint-lists.py.  I made it super picky; it will even complain about "2017-3-1" instead of "2017-03-01".  This is to make life as easy as possible for people trying to process the CSV files in the future.  Whenever it was unclear what a date should have been, I used the Git commit date as a guideline.

N.B. the changes to lint-lists.py also include two unrelated bugfixes: the initialization of 'date_added' at the beginning of main is removed because it's a dead store, and an initialization "idx = -1" is added ahead of the per-file loop to prevent the script from crashing on empty test lists (e.g. lists/ni.csv).


scripts/lint-lists.py
```diff
@@ -21,6 +21,20 @@
 LEGACY_CATEGORY_CODES = "00-LEGEND-category_codes.csv"
 COUNTRY_CODES = "00-LEGEND-country_codes.csv"
 
+def is_valid_date(date):
+    try:
+        y, m, d = date.split('-')
```
hellais: Why not just use: `datetime.datetime.strptime(date, "%Y-%M-%d")`?

hellais:
Good, stuff thanks for doing this!

I left just a small comment. Perhaps it's easier to just use the `datetime.strptime` parser (though I can't actually think of an issue with your approach).

scripts/lint-lists.py
zackw: Mainly because it returns a datetime.date rather than a datetime.datetime.  There's no datetime.date.strptime method.  Also, I wasn't sure whether you needed this code to work with really ridiculously old Python (datetime.datetime.strptime was added in 2.5).

I'm happy to change it to use strptime if you would prefer.  Please let me know, then I'll rebase it.

hellais: I see. Yeah legacy python support is not a concern as this is just meant to run on the CI. I would change it to using standard python `datetime` parsing functions to make it more robust to other forms of invalid dates (ex. `2018-02-31`).

zackw:
I have updated the branch with the conflicts fixed and `is_valid_date` changed to use `datetime.strptime`.

hellais:
Excellent, thank you so much!

PR #328, opened 2018-04-10
This file lists the URLs to test from Ecuador.

iapazmino:
This is a starting set just to begin gathering data from Ecuador.


hellais:
Thanks for putting this together!

It looks good. Merging.

PR #332, opened 2018-04-13
Initial list for Uruguay. Created after court order to block Roja Directa

ilv:
Created an initial list of websites for Uruguay. I hope is not too small.


hellais:
Great work. Thanks!

PR #353, opened 2018-06-08
Added news sites to global list

agrabeli:
I have added some news sites (provided by RSF) to the global test list.


agrabeli:

hellais:
Fixed.

agrabeli:
@hellais @sneft The "checks have failed" because there are duplicate URLs between this and the US test list. These URLs though are prominent US news websites, which RSF added so that they can get tested internationally (since they have been blocked in various countries). It therefore seems reasonable that these URLs should be included in the global and US test lists, but it's not possible to merge this PR with these duplicates. How do you propose we go about this?

agrabeli:
@hellais @sneft I tried to resolve this problem by deleting these 6 URLs from the US test list. See: https://github.com/citizenlab/test-lists/pull/456 This way, these (internationally relevant) URLs are only included in the global test list, and they'll be tested in the US anyway (since OONI Probe always tests the URLs in the global list, regardless of country). I think merging the #456 PR is probably a requirement to merging this one.

agrabeli:
@hellais @sneft Perhaps this is now ready to be merged?

PR #383, opened 2018-08-17
Add URLs to China list

ahounsel:
Added some new China-specific URLs to the test list for China. Most of these URLs belong to news organizations that publish political dissent.


hellais:
lgtm, thanks!

PR #386, opened 2018-09-03
Updated CD test list

agrabeli:
I have updated the CD test list with URLs provided by community members.


agrabeli:

PR #388, opened 2018-09-05
Updated the JO test list

agrabeli:
As part of the updates to the JO test list, I have:

1. Deleted URLs whose domains have expired, been squatted, or are otherwise no longer operational
2. Added URLs based on research (in an attempt to identify relevant & interesting URLs to test)


agrabeli:

PR #389, opened 2018-09-13
Updated CN,EG,ET,RU,TR,IR,Global test lists

agrabeli:
Sprinkled media sites across test lists.


hellais:
@sneft can you look into merging this PR together with the other ones I assigned to you.

PR #391, opened 2018-09-18
Updated UG & global test lists

agrabeli:
This PR adds URLs to the UG & global test lists, as provided by Uganda's DefendDefenders.


agrabeli:
@sneft @hellais This PR is ready for being merged.

PR #401, opened 2018-10-12
Remove escape characters from URLs in india list

hellais:
We noticed when ingesting the lists into a DB that some URLs contained escape characters.


lists/in.csv
```diff
@@ -391,7 +391,6 @@ http://www.datedosti.com,DATE,Online Dating,2016-06-14,OONI,
 http://www.deepalaya.org/,HUMR,Human Rights Issues,2016-06-14,OONI,
 http://www.defence.pk/forums/central-south-asia/203273-mim-chief-urges-muslims-help-assam-riots-victims.html,MILX,Terrorism and Militants,2014-04-15,citizenlab,
 http://www.defence.pk/forums/indian-defence/203141-muslims-face-executions-indias-assam-region.html,MILX,Terrorism and Militants,2014-04-15,citizenlab,
-http://www.defence.pk/forums/world-affairs/195202-massarce-muslims-burma-myanmar-\state-rohingyas-community-massacre.html,NEWS,News Media,2014-04-15,citizenlab,
 http://www.defence.pk/forums/world-affairs/195202-massarce-muslims-burma-myanmar-state-rohingyas-community-massacre-2.html,REL,Religion,2014-04-15,citizenlab,
```
hellais: This is a dupe as the URL with the `-2` suffix redirects to the same page as the one without it.

lists/in.csv
```diff
@@ -427,7 +426,7 @@ https://www.facebook.com/groups/346504835433106/?ref=ts,REL,Religion,2014-04-15,
 https://www.facebook.com/groups/349745378434446/,REL,Religion,2014-04-15,citizenlab,
 http://www.facebook.com/groups/410690962321650,REL,Religion,2014-04-15,citizenlab,
 http://www.facebook.com/groups/421881071186596/?ref=ts,REL,Religion,2014-04-15,citizenlab,
-http://www.facebook.com/groups/43700222634264\ad\ad3/?ref=ts,REL,Religion,2014-04-15,citizenlab,
+http://www.facebook.com/groups/43700222634264,REL,Religion,2014-04-15,citizenlab,
```
darkk: That _probably_ should be https://www.facebook.com/groups/437002226342643
At least that's alive link. I'm not sure what `\ad` is.

darkk:
:+1:

PR #402, opened 2018-10-13
added https://pr0gramm.com

Strobotron:
This controversial (often described as right wing) german image board seems to be delisted from google searches within Germany and seems to be inaccessible from the uk


muellermartin:
This seems to be superseded by #463 and can probably be closed.

PR #403, opened 2018-10-15
Contribute url to global list

iliasbartolini:


hellais:
There is a conflict with the base branch, so it needs conflict resolution.

hellais:
Thanks!

PR #405, opened 2018-10-16
Add Cloudflare and Quad9 DNS over HTTPS servers.

R6CG:
These complement the existing https://dns.google.com/. For these new
ones, I added a path and query string that actually does a lookup for
www.example.com, rather than pointing at their landing page.

Besides their memorable IPv4 address, these services also have IPv6
addresses. But as far as I can tell, there are no existing URLs with
literal IPv6 addresses. Additional addresses, not added by this patch,
are:

Quad9
	149.112.112.112
	2620:fe::fe
	2620:fe::9

Cloudflare
	2606:4700:4700::1111
	2606:4700:4700::1001

In lookup up these providers, I also found
https://cleanbrowsing.org/dnsoverhttps operating a public DNS over HTTPS
server.


hellais:
This seems like a good thing to have in the global list. Merging.

PR #406, opened 2018-11-09
Update gr.csv list

anadahz:


anadahz:
Updating the Greek test-list.

lists/gr.csv
```diff
@@ -38,3 +38,8 @@ https://www.migrant.gr/,HUMR,Human Rights Issues,2018-09-20,OONI,
 http://www.efsyn.gr/,NEWS,News Media,2018-09-20,OONI,
 http://astegoi.gr/,HUMR,Human Rights Issues,2018-09-20,OONI,
 https://www.amnesty.gr/,HUMR,Human Rights Issues,2018-09-20,OONI,
+http://www.subtitles.gr,SRCH,Search Engines,2018-11-09,Anonymous,
+http://www.subztv.club/,CULTR,Culture,2018-11-09,Anonymous,
```
hellais: I believe these two should be categorised as FILE.

lists/gr.csv
```diff
@@ -38,3 +38,8 @@ https://www.migrant.gr/,HUMR,Human Rights Issues,2018-09-20,OONI,
...
+http://www.yts.am/,FILE,File-sharing,2018-11-09,Anonymous,
+http://www.loosecannonholdem.com/,MISC,Miscelaneous content,2018-11-09,Anonymous,
```
hellais: This should probably go under control content (we try to avoid using MISC for all new PRs and it was only added due to some incompatibility in the new vs old category codes).

lists/gr.csv
anadahz:
```
File-sharing 	FILE 	P2P SFTWR WRZ 	Sites and tools used to share files, including cloud-based file storage, torrents and P2P file-sharing tools.
```
I don't think that they fit into the FILE category, these sites are not torrent p2p or cloud-based file storage sites, if we want to add them into this category perhaps we should also update the category description.

lists/gr.csv
anadahz: Thanks for the review, updated the category.

anadahz:
Update PR based on @hellais comments.

anadahz:
Reminder @hellais anything else missing from this PR?

PR #407, opened 2018-11-16
Cryptocurrency monitoring

willscott:
This adds a few of the major cryptocurrency exchanges, and bootstrap / entry points to bitcoin and ethereum networks.

* The bitcoin entry points use DNS lookups from hard-coded dns names [in the source](https://github.com/bitcoin/bitcoin/blob/863e995b79ec388bf292d80f181912d01e20e2e5/src/net.cpp#L1128)
* In addition, DNS and IP based reachability would complement that, but is out of scope for this list.
* Infura is the most popular web gateway for submitting transactions into ethereum, and is a reasonable proxy for whether there is interference with that network.
* The others are a global selection of the largest exchanges listed on [wikipedia](https://en.wikipedia.org/wiki/Cryptocurrency_exchange)

There have already been reports of cryptocurrencies being blocked by countries. Seems worth tracking.

### DIFF
--- a/lists/global.csv
+++ b/lists/global.csv
@@ -31,6 +31,7 @@ http://beeg.com,PORN,Pornography,2014-04-15,citizenlab,Updated by OONI on 2017-0
 https://beyondexgay.com/,LGBT,LGBT,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 https://bisexual.org/,LGBT,LGBT,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 http://bitcoin.org,COMM,E-commerce,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
+http://bitcoin.sipa.be/,COMM,E-commerce,2018-11-15,blockchain,Initial tracking of cryptocurrencies on 2018-11-15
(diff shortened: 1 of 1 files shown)


lists/global.csv
```diff
@@ -67,6 +68,7 @@ http://creativecommons.org,NEWS,News Media,2014-04-15,citizenlab,Updated by OONI
 https://cryptome.org,NEWS,News Media,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 http://cultdeadcow.com,HACK,Hacking Tools,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 https://cyber.harvard.edu/,NEWS,News Media,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
+https://dashjr.org/,COMM,E-commerce,2018-11-15,blockchain,Initial tracking of cryptocurrencies on 2018-11-15
```
hellais: These are bootstrapping addresses hardcoded into the bitcoin source code: https://github.com/bitcoin/bitcoin/blob/863e995b79ec388bf292d80f181912d01e20e2e5/src/net.cpp#L1128

hellais:
We should add an extra comment explaining why some of these addresses are relevant to cryptocurrencies, but otherwise it's good to go.

hellais:
Thanks!

PR #408, opened 2018-11-24
Add Thai political party urls

bact:
- add Thai political party urls
- few news media and political criticism sites


hellais:
Thanks for this! It looks good. I am merging.

Issue #400, opened 2018-10-11
Extra consistency checks on category codes and URLs


hellais:
We have run into the issue, when using the lists in OONI, that some country lists present the following problems:

1. The same URL present in different country specific lists, presents a different category code

ex.
id.csv:http://denypagetests.netsweeper.com,NEWS,News Media,2014-04-15,citizenlab,
kw.csv:http://denypagetests.netsweeper.com,CTRL,Control content,2014-04-15,citizenlab,

2. The same URL is present in both the global and the country specific list

ex.
global.csv:http://www.crazyshit.com,PORN,Pornography,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
sg.csv:http://www.crazyshit.com,NEWS,News Media,2014-04-15,citizenlab,

We should add checks to the lint-lists.py script that checks if:

1. There are inconsistencies in category codes across lists
2. If a URL is present in the global list it should not also be present in the country specific list

On this second point I would like to hear from @sneft and others to know if this is reasonable or if it's maybe just a OONI specific usage of the lists.

darkk:
Check №2 is a bit tricky for `cis.csv`. Should `cis.csv` be treated in a same way as `global.csv` for corresponding countries? What definition of CIS should it use? E.g. should it include Georgia?

hellais:
At OONI we don't actually use `cis.csv` at all and that country list has not been updated in a pretty long while. I would go to the extent of suggesting we remove it or move it to another directory.

sneft:
For point 1, I have no doubt there are a number of these inconsistencies. We tried to fix these as we encountered them but haven't ever made a systematic effort to clean them.

For point 2, I agree that if a URL is present on the global list it should not be on a local list. Our old testing system flagged when you attempted to upload a local list with a URL duplicated in the global list. Our logic was that the global and local lists are meant to be run as a single unit, so we wanted to avoid duplication. I know we had some cases where this was inconvenient (e.g. wanting to test a very narrow sample in a bandwidth-limited place) but to my knowledge OONI is flexible enough to better accommodate custom lists for special circumstances.

(This requirement does add a small burden of labour on list compilers, as in my experience the average person compiling a local list will reasonably (and often appropriately) add certain URLs that are duplicated in the global list. Perhaps this is just a matter of good documentation and instructions to list compilers.)

PR #409, opened 2018-11-30
removes dupes between local list and global list and adds checking

jakubd:
Removes duplicates between local list and global list (there were 194 cases where local list duplicated entries in the global list)

Also added a check in lint-lists.py that checks for this.  Does not address the cis.csv case.

see #400

### DIFF
--- a/lists/global.csv
+++ b/lists/global.csv
@@ -1133,9 +1133,12 @@ https://pastebin.com/,HOST,Hosting and Blogging Platforms,2017-02-12,Anonymous,O
 https://yandex.com/,SRCH,Search Engines,2018-02-23,igorv,
 https://protonmail.com/,HACK,Hacking Tools,2018-03-19,OONI,
 https://getoutline.org/,ANON,Anonymization and circumvention tools,2018-04-06,Duplika,
+http://fteproxy.org/,ANON,Anonymization and circumvention tools,2018-05-21,Community member,
+https://www.thegeekdiary.com/,HACK,Hacking Tools,2018-05-21,Community member,
+https://dns.google.com/,HOST,Hosting and Blogging Platforms,2018-05-02,OONI,
+https://soundcloud.com/,CULTR,Culture,2018-07-23,OONI,
 https://signal.org/,COMT,Communication Tools,2018-04-19,OONI,
 https://www.tunnelbear.com/,ANON,Anonymization and circumvention tools,2018-04-19,OONI,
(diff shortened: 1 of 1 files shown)


hellais:
@jakubd thanks for putting this together! I think it's a great idea.

Would you mind splitting the commits making changes to the `.csv` files and the `lint-list.py` scripts? This will make it easier to track the changes in the future.

scripts/lint-lists.py
```diff
@@ -92,6 +95,15 @@ def load_categories(path, get_description_code=get_new_description_code):
...
     return code_map
 
+def load_global_list(path):
+    check_list = {}
+    with open(path, 'rb') as in_file:
+        reader = csv.reader(in_file, delimiter=',')
+        for idx, row in enumerate(reader):
+            if idx != 0 and (len(row) == 6):
+                check_list[row[0]] = row[1]
+    return check_list
```
hellais: You can probably just use a `set` here instead of `dict`, since you don't need the value.

scripts/lint-lists.py
```diff
@@ -109,6 +121,8 @@ def main(source='OONI', notes='', legacy=False, fix_duplicates=False):
         )
     header = ['url', 'category_code', 'category_description',
               'date_added', 'source', 'notes']
+    # preload the global list to check against looking for dupes
+    check_against_global = load_global_list(glob(os.path.join(lists_path, "global.csv"))[0])
```
hellais: If you make the above change to use a `set`, I would suggest renaming this to `global_urls_bag` for consistency with the `urls_bag` below.

hellais: Also if you have passing the path the list (which is precise) you don't need to use `glob` in front of this, but you can just say: `load_global_list(os.path.join(lists_path, "global.csv"))`

scripts/lint-lists.py
```diff
@@ -135,6 +149,11 @@ def main(source='OONI', notes='', legacy=False, fix_duplicates=False):
                     )
                 url = url.strip().lower()
                 canonical_url = url
+                if os.path.basename(csv_path) != "global.csv":
+                    if canonical_url in check_against_global:
+                        errors.append(
+                            DuplicateURLWithGlobalList(canonical_url, csv_path, idx+2)
+                        )
```
hellais: I would suggest doing this check against the "canonical_url" and storing in the `global_urls_bag` the canonical representation of the url (this way we will detect dupes when we have two URLs that are `http://foo.com/` & `http://foo.com`).

PR #411, opened 2018-12-03
removes dupes between local list and global list and adds checking

jakubd:
integrated changes proposed by @hellais in #409 just lack the git fu to split commits retroactively 

---

Removes duplicates between local list and global list (there were 194 cases where local list duplicated entries in the global list)

Also added a check in lint-lists.py that checks for this. Does not address the cis.csv case.

see issue #400


hellais:
This looks good to me. Thanks for putting it together!

PR #422, opened 2018-12-27
added new organizations and media

VirtualRoad:
added new organizations and media in connection with the upcoming elections


hellais:
lgtm!

PR #428, opened 2019-01-31
Updated MX test list

agrabeli:
This update includes new URLs added to the MX test list. 

As part of the review, I opened every URL that was previously included in order to check whether they're still functional, relevant, etc. This resulted in the deletion of many URLs that have been squatted or are otherwise no longer functional. I made some changes to URLs (those that have changed over time) and changed some category codes (for example, once the FEXP category code was dropped, many sites were re-categorized to NEWS, but aren't).

Hopefully these proposed updates to the MX test list will render better measurements!


agrabeli:
@hellais @sneft Please merge this? :)

sneft:
Thank you! This was likely a lot of work so thanks for doing it. Merging.

PR #433, opened 2019-02-19
Update Wikipedia-specific URLs

azadi:
This commit updates the Wikipedia URLs in the global and local lists.
Notable changes include:

- moving ar.wikipedia.org and wikidata.org from the local lists to the
global lists (and removing their entries in the former)

- adding the mobile version of the top domains (.m) to both the lists to
reflect the large percentage of pageviews from mobile devices

- removing the HTTP URLs as Wikipedia has HTTPS enabled and uses HSTS


lists/global.csv
```diff
@@ -59,7 +59,6 @@ https://cites.org/,ENV,Environment,2014-04-15,citizenlab,Updated by OONI on 2017
 https://citizenlab.ca/,HUMR,Human Rights Issues,2014-04-15,citizenlab,Updated by citizenlab on 2017-11-09
 http://cocaine.org/,ALDR,Alcohol & Drugs,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 http://common-fund.org/,ECON,Economics,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
-http://commons.wikimedia.org/,GRP,Social Networking,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
```
hellais: This is dropped as we already have the https version of it in the list.

hellais:
Thank you so much for putting the work into this and also to the rest of the folks at wikimedia for helping us improve the testing of such an important service!

This pull request looks good to merge 🐳

PR #434, opened 2019-02-20
Updated SN test list

agrabeli:
The SN test list has been updated based on URLs provided by Senegal's ASUTIC.


hellais:
Merging as it's urgent for the elections.

PR #444, opened 2019-03-08
add riseup black vpn service

kalikaneko:


hellais:
It needs some conflicts to be resolved

kalikaneko:
@hellais rebased

anadahz:
@kalikaneko Why only in China, Iran and India?
Isn't useful to test Riseup's VPN service worldwide?

kalikaneko:
@anadahz I'm not too familiar about the process to include uris in the list, so I added just to the few places where I've heard it was being problematic. But if it's not a problem I can add it worldwide.

anadahz:
@kalikaneko Just add the URL to https://github.com/citizenlab/test-lists/blob/master/lists/global.csv .
The URL will be tested by every probe on the planet. :)

Thank you for contributing URLs to test-lists.

hellais:
The homepage of https://black.riseup.net/ says:
> This service is being discontinued, in favor of RiseupVPN.

Should we perhaps be testing a different domain?

I agree that it would perhaps be beneficial to add a service of this sorts to the global list.

hellais:
I see that the client is written in golang (after a quick inspection of https://0xacab.org/leap/bitmask-vpn), so perhaps it's also worth considering possibly writing an OONI Probe test for it.
With @bassosimone we are working on making the measurement-kit engine work nicely with golang code and have tried this out for another circumvention toll.

kalikaneko:
rebased again, adding it to global.

> Should we perhaps be testing a different domain?
there was a service with username registration, and that's what riseup is deprecating in favor of RiseupVPN (a branded build of the generic go client). But the black server still is relevant, since it's the entrypoint for bootstrapping the whole configuration. Requests after the initial bootstrapping go to an api server with self-signed certificates, so I suspect the output of adding that one to the uri list will not be helpful.

@hellais I'll investigate how to write a probe test for the client, thanks

RandomGuy48:
#972

PR #445, opened 2019-03-18
Tedic's proposed list

lupa18:
Hello, Tedic propose this Paraguayan list.


agrabeli:
Awesome, thank you @lupa18 !

Before this gets merged, I'd propose the following minor changes:

* Please re-categorize the URLs categorized as MISC with some other category (https://github.com/citizenlab/test-lists/blob/master/lists/00-LEGEND-new_category_codes.csv). We only added MISC as a temporary category to categorize un-categorized URLs (tested independently via the apps), but it would be preferable if every URL has one of the 30 other categories.

* Please add TEDIC in the second last column ("source")

Thank you! :)

agrabeli:
Thank you @lupa18 !

@hellais @sneft Please consider merging? :)

PR #446, opened 2019-03-19
Add Tor directory authorities server descriptors

anadahz:


sssics:
Thanks @anadahz 

This list is very important for countries like Mexico.

nvjacobo:
Thanks @anadahz 

This list of IP addresses is very important. For example, in Mexico there have been incidents related to these IP addresses.

hellais:
This is great, however I think it's most suited to be part of https://github.com/OpenObservatory/ooni-resources/blob/master/bridge_reachability/tor-bridges-ip-port.csv (where it already is present) as this is not a website.

anadahz:
I think we can add the Tor dirauths as a "special case" similar to the Cloudflare DNS over HTTPS entries in the test-list:  https://github.com/citizenlab/test-lists/blob/master/lists/global.csv#L1158

agrabeli:
@hellais While it's true that test lists are primarily meant to include websites, perhaps we can make an exception here (similarly to the Cloudflare example pointed out by @anadahz) given that the OONI Probe tests for Tor bridge reachability are currently only available via the legacy version of OONI Probe (which is not run much at the moment around the world). 

@sneft thoughts? How do you propose we proceed with this PR?

agrabeli:
@hellais @sneft How should we proceed with this PR?

PR #452, opened 2019-03-28
Kiwifarms and 8chan to global list

fifteenthcommotion:
Add kiwifarms.net and 8ch.net as Social Networking sites.


lists/global.csv
```diff
@@ -325,6 +326,7 @@ http://www.4online-gambling.com/,GMB,Gambling,2014-04-15,citizenlab,Updated by O
 http://www.4shared.com/,FILE,File-sharing,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 http://www.888casino.com/,GMB,Gambling,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 http://www.89.com/,PORN,Pornography,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
+http://8ch.net/,GRP,Social Networking,2019-03-27,citizenlab,
```
hellais: The contributor should not be citizenlab in this case

hellais:
I believe the reason these are being added is that these sites were being interfered with following the terrorist attacks in NZ. I am not 100% sure we want to be adding them and would like some input from @sneft on this.

fifteenthcommotion:
Should I change contributor to be fifteenthcommotion?

agrabeli:
^^ @sneft

PR #454, opened 2019-04-06
Added thepiratebay.org

Bodup:
thepiratebay.org is a known censored domain in The Netherlands.


hellais:
thepiratebay.org is already part of the global list so there is no need to add it to the country list.

PR #455, opened 2019-04-09
Updated IN test list

agrabeli:
India's 2019 general elections start in 2 days so I updated the IN test list to include a number of missing political party sites. I tried to include as many of those participating in the 2019 elections as possible, but not all political parties seem to have functional websites (and some don't appear to have websites at all). I also added several election-specific websites.

@hellais @sneft I'd appreciate it if you could please review & merge in time for the elections. Thanks for your time!


hellais:

PR #87, opened 2016-10-04
Create test list for Uruguay

hellais:
Creating this pull request for the addition of the test list from @asterismo.

Original comment from his commit is:

> These are a collection of URLs worth testing for country code UY

In order to make this mergable the following fixes need to be made:
- [ ] The columns need to be 6 (url,category_code,category_description,date_added,source,notes). What is now the url_description column should probably go inside of the notes section.
- [ ] Change the blogspot URLs to be the canonical .com version (I think it's more likely that these are blocked rather than their `.uy` equivalent). This is a non blocker though.

PR #457, opened 2019-04-18
Updating websites from Uruguay to test for censorship.

gabelula:
It is adding news website to the list and adding some of the websites from https://github.com/citizenlab/test-lists/pull/87/ .


gabelula:
CI says that many of the urls are invalid... but they seem fine to me.

hellais:
They are missing the trailing `/` at the end of them. I agree that the error message could be a bit more descriptive.

I ran:
```
python scripts/lint-lists.py --fix-slash lists
```

To fix the issue with the slashes and pushed to your branch.

PR #462, opened 2019-05-02
Update lists/cn.csv to reflect recent censorship events

azadi:
Update lists/cn.csv to reflect recent censorship events

Add zh.wikinews.org to the list


hellais:
lgtm

PR #463, opened 2019-05-08
Added Pr0gramm.com

Strobotron:
Pr0gramm is a controversial german image board that is often described as right wing or alt right. It does not seem to appear in google searches


hellais:
Needs conflict resolution

muellermartin:
Hm, according to the CI, this is the error:

    lists/de.csv (line 10): Invalid URL "https://pr0gramm.com"

The URL looks good. Maybe it's a false alert?

agrabeli:
@muellermartin I think it says "invalid URL" because the "source" column is empty (and something needs to be included there, even if it just says "anonymous") in order for the PR to pass the checks.

agrabeli:
Aha! It seems that the "invalid URL" error was mainly because the slash was missing at the end of the URL. Just fixed this PR. Merging now.

muellermartin:
> @muellermartin I think it says "invalid URL" because the "source" column is empty (and something needs to be included there, even if it just says "anonymous") in order for the PR to pass the checks.

Hm, I've submitted PRs with empty source columns (e.g. see #491) and that did not trigger the CI warning. I remember that I've tested it locally with the `lint-lists.py` script and adding a trailing slash did make a difference, but I'm not sure...

PR #465, opened 2019-05-09
Update cu.csv

barbaramaseda:


hellais:
Thanks for putting this together!

I made a couple of fixes to the URLs and it looks good to merge to me.

PR #469, opened 2019-05-30
Update global.csv

spmedia:
om.org (Operation Mobilization) has updated to use -> https://www.omusa.org/

This PR updates https://www.om.org/ to https://www.omusa.org/

You can verify this url update at:
- https://www.facebook.com/omusa.org
- https://twitter.com/omusa

Visiting https://www.om.org/ via the OONI app on Android throws an error (!) when it attempts to scan it and throwing a false positive as "Likely blocked" because they have updated to a new URL.

Thanks

### DIFF
--- a/lists/global.csv
+++ b/lists/global.csv
@@ -835,7 +835,7 @@ http://www.oic-oci.org/,REL,Religion,2014-04-15,citizenlab,Updated by OONI on 20
 http://www.oicc.org/,REL,Religion,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 http://www.oikoumene.org/,REL,Religion,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 https://www.okcupid.com/,DATE,Online Dating,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
-https://www.om.org/,REL,Religion,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
+https://www.omusa.org/,REL,Religion,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 http://www.omct.org/,HUMR,Human Rights Issues,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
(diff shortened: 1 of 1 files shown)


lists/global.csv
```diff
@@ -835,7 +835,7 @@ http://www.oic-oci.org/,REL,Religion,2014-04-15,citizenlab,Updated by OONI on 20
 http://www.oicc.org/,REL,Religion,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 http://www.oikoumene.org/,REL,Religion,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 https://www.okcupid.com/,DATE,Online Dating,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
-https://www.om.org/,REL,Religion,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
+https://www.omusa.org/,REL,Religion,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
```
hellais: I still see content on https://www.om.org/. I think omusa.org is the US chapter of OM similarly to how there is a Canadian chapter on omcanada.org, see: https://twitter.com/OM_Canada.

I suggest you add this site to the US testing list or to the country lists.

muellermartin:
A @hellais pointed out, the URL https://www.omusa.org/ seems to be the US chapter of OM in the US. The linked websites provides no proof that the original URL was updated. In contrast to that there is https://twitter.com/OMinternational which still points to http://www.om.org/ and there are a lot of other Twitter accounts for chapters in other nations, that point to their own URLs (or to om.org):

- https://twitter.com/om_br → om.org.br
- https://twitter.com/OMIreland → ie.om.org
- https://twitter.com/OMChile → om.org
- https://twitter.com/OM_Australia → om.org.au
- https://twitter.com/OMKAMusa → omkam.org
- https://twitter.com/omchina → om-china.org

Therefore I would opt for closing this issue and instead propose to additionally include the other URLs of the national chapters.

spmedia:
Thanks for the comments, I agree with you @muellermartin .

PR #471, opened 2019-05-30
Update cn.csv

mdrights:
A blogging platform (on social issues) is added (already blocked);
A distributed communication tool service is added (may be at risk to be blocked).


hellais:
lgtm

PR #473, opened 2019-06-07
Update Wikisource URLs

azadi:
Update the French-speaking countries list with fr.wikisource.org and fr.m.wikisource.org, and Russian-speaking countries with ru.wikisource.org and ru.m.wikisource.org


hellais:
lgtm

PR #484, opened 2019-07-17
Add gitlab bintray twimg to global.csv

xhdix:
Added various twitter related URLs due to the recent Twitter censorship method in Indonesia.


lists/global.csv
```diff
@@ -1234,3 +1234,10 @@ https://pt.m.wikipedia.org/,CULTR,Culture,2019-02-19,Wikimedia,URLs provided by
 https://mitm.watch/,ANON,Anonymization and circumvention tools,2019-03-20,magma,This site detects encryption-removing MITM using logic from NDSS17 interception paper
 https://credibilitycoalition.org/,CULTR,Culture,2019-04-19,OONI,Reportedly inaccessible on WE in Egypt (but probably worth testing globally)
 https://black.riseup.net/,ANON,Anonymization and circumvention tools,2019-04-26,LEAP,
+https://gitlab.com/,HOST,Hosting and Blogging Platforms,2019-07-17,chelchela,
+https://termux.com/,HOST,Hosting and Blogging Platforms,2019-07-17,chelchela,
+https://dl.bintray.com/termux/,HOST,Hosting and Blogging Platforms,2019-07-17,chelchela,
```
hellais: Why termux specifically and not the root of bintray.com?

lists/global.csv
```diff
@@ -1234,3 +1234,10 @@ https://pt.m.wikipedia.org/,CULTR,Culture,2019-02-19,Wikimedia,URLs provided by
 https://mitm.watch/,ANON,Anonymization and circumvention tools,2019-03-20,magma,This site detects encryption-removing MITM using logic from NDSS17 interception paper
 https://credibilitycoalition.org/,CULTR,Culture,2019-04-19,OONI,Reportedly inaccessible on WE in Egypt (but probably worth testing globally)
 https://black.riseup.net/,ANON,Anonymization and circumvention tools,2019-04-26,LEAP,
+https://gitlab.com/,HOST,Hosting and Blogging Platforms,2019-07-17,chelchela,
+https://termux.com/,HOST,Hosting and Blogging Platforms,2019-07-17,chelchela,
```
hellais: Is there a specific reason to test the homepage of this software?

lists/global.csv
xhdix: At the first start of the Termux app, it must communicate with its site, otherwise the application will remain in loading mode. To be sure, I'll re-analyze the app.

xhdix: I made a mistake
The termux.org and termux.net addresses are used :
https://github.com/termux/termux-app/blob/370ac2bd8911b5632933975300e779f722ad03af/app/src/main/java/com/termux/app/TermuxInstaller.java#L174-L175

The following URLs should be tested for proper analysis:
https://termux.org/bootstrap-arm.zip
https://termux.net/bootstrap/bootstrap-arm.zip

lists/global.csv
xhdix: At the first start of the Termux app, it must communicate with its site, otherwise the application will remain in loading mode.

I made a mistake
The termux.org and termux.net addresses are used :
https://github.com/termux/termux-app/blob/370ac2bd8911b5632933975300e779f722ad03af/app/src/main/java/com/termux/app/TermuxInstaller.java#L174-L175

The following URLs should be tested for proper analysis:
https://termux.org/bootstrap-arm.zip
https://termux.net/bootstrap/bootstrap-arm.zip

lists/global.csv
xhdix: Just for having status code: 200 at dl.bintary.com which I think is more important than bintray.com.

lists/global.csv
```diff
@@ -1234,3 +1234,11 @@ https://pt.m.wikipedia.org/,CULTR,Culture,2019-02-19,Wikimedia,URLs provided by
 https://mitm.watch/,ANON,Anonymization and circumvention tools,2019-03-20,magma,This site detects encryption-removing MITM using logic from NDSS17 interception paper
 https://credibilitycoalition.org/,CULTR,Culture,2019-04-19,OONI,Reportedly inaccessible on WE in Egypt (but probably worth testing globally)
 https://black.riseup.net/,ANON,Anonymization and circumvention tools,2019-04-26,LEAP,
+https://gitlab.com/,HOST,Hosting and Blogging Platforms,2019-07-17,chelchela,
+https://termux.org/bootstrap-x86_64.zip,HOST,Hosting and Blogging Platforms,2019-07-17,chelchela,
+https://termux.net/bootstrap/bootstrap-arm.zip,HOST,Hosting and Blogging Platforms,2019-07-17,chelchela,
```
hellais: These are 5MB files which if we include as part of the global list will be fetched by every client.

Has there been an instance of this termux app being blocked in some country? Perhaps we should add these URLs to the country specific list?

lists/global.csv
hellais: If that is the goal, perhaps we can use the more "neutral" URL for bintray: https://dl.bintray.com/bintray/.

Is this in response to some incident of blocking of bintray.com?

lists/global.csv
xhdix: It was blocked in Iran until last month. But now on an ISP I didn't see any blockage. (Except for a lot of problems with their https communications)
It is better to test it globally since it is a world famous application. Or, if `status code: 404` is okay, I'll change it to `https://termux.net/` and `https://termux.org/` .

lists/global.csv
xhdix: Does this URL respond to you?!

hellais: Yeah you are right it doesn't respond reliably. How about https://dl.bintray.com/jfrog/xray/?

xhdix: From the Netherlands:
image

lists/global.csv
hellais: How about we start off by adding it to the Iran list?

hellais: Let's change these to the index (even if they 404) and add them inside of the Iran test list.

lists/global.csv
hellais: Yeah you are right, the `/bintray` URL does not work well.

What about `/jfrog/xray` (jfrog is the company behind bintray)?

lists/global.csv
xhdix: OK.

lists/global.csv
xhdix: done

hellais:
Resolved the conflict and it looks good to merge.

Thanks for working on this!

lists/global.csv
anadahz: Does it makes sense to test 3 HTTPS URL of the same domain?
The 2 files of `termux.org` URLs  account for ~20M (5.8M bootstrap-arm.zip,  14M bootstrap-x86_64-v13.zip).

xhdix: @anadahz The decision changed: https://github.com/citizenlab/test-lists/pull/490

Issue #16, opened 2015-06-29
Make list of government issued blocklists


hellais:
# Add links into the comments below!

Are you aware of a government issued blocklist that is available for download?

Then add a link pointing to where it can be downloaded (even better if you specify if it's legal or not to mirror/publish it) and what country to applies to.

If you want to be :+1: :100: then you should write the parser code and submit a pull request ;)

jakubd:
Two that jump out in my mind are:

Indonesia - http://trustpositif.kominfo.go.id/
Russia - http://eais.rkn.gov.ru/  or the less official but very reliable antizapret

Some countries will need to be scraped often so we can easily see when URLs are added/removed.  We would need to stage a machine to check and update the repo when this happens.

arglebarf:
There's http://engelliweb.com/ for Turkey. 

Not a gov site, but a volunteer.

graphiclunarkid:
There's a bunch of information for the UK, and limited information on other countries, on the Error 451 wiki: http://wiki.451unavailable.org.uk/wiki/Main_Page

It'd be wonderful if you felt like updating the site with the information from this thread too :smile:

hellais:
@arglebarf @graphiclunarkid could you submit a PR with the URLs inside of that list?

You can see how to properly format and structure the parsers in this directory: https://github.com/citizenlab/test-lists/tree/master/lib/lists/official/it

darkk:
RU: http://reestr.rublacklist.net/api/current (that's unofficial API)

kaerumy:
List of escort sites blocked in india:

http://cis-india.org/internet-governance/blog/list-of-blocked-escort-service-websites

```
www.sterlingbioscience.com
rawpoint.biz
www.onemillionbabes.com
www.mumbaihotcollection.in
simranoberoi.in
rubinakapoor.biz
talita.biz
www.mumbaiescortsagency.net
www.mumbaifunclubs.com
www.alishajain.co.in
www.ankitatalwar.co.in
https://www.jennyarora.ind.in
www.riya-kapoor.com
shneha.in
missinimi.in
www.mumbaiglamour.in
kalyn.in
www.saumyagiri.co.in/city/mumbai/
bookerotic.com
www.divyamalik.in
www.suhanisharma.co.in
www.ruhi.biz
umbaiqueens.in
www.aliyaghosh.com
priyasen.in
www.highprofilemumbaiescorts.co.in
charmingmumbai.com
www.poojamehata.in
kiiran.in/
[...]

PR #489, opened 2019-08-06
Add GitHub government takedown requests

muellermartin:
Manually added by looking at the official GitHub repository for
government takedown requests at https://github.com/github/gov-takedowns

The URLs were mostly retrieved via following command (with GNU grep)
from within the cloned GitHub repository:

    grep -roP --exclude-dir='.?*' 'https?://.*? ' . | sort | uniq | grep -v http://eais.rkn.gov.ru

URLs which lead to HTTP 404 errors, and thus cannot be categorized properly, were omitted (currently 5).

Although the Russian takedown requests announce that the offending URLs will be added to the "Unified register of the domain names, website references and network addresses that allow identifying websites containing information circulation of which is forbidden in the Russian Federation", only a few are listed there (mostly the newer ones from 2019). Therefore this PR is somewhat related to issue #16 This can be checked here: http://eais.rkn.gov.ru/en/

The mapping to the categories are highly subjective, therefore feedback is more than welcome!

### DIFF
--- a/lists/ru.csv
+++ b/lists/ru.csv
@@ -311,3 +311,34 @@ https://www.svoboda.org/,NEWS,News Media,2018-09-13,OONI,
 https://ru.m.wikipedia.org/,CULTR,Culture,2019-02-19,Wikimedia,URLs provided by wikipedia based on the most popular language domains per country
 https://ru.wikisource.org/,CULTR,Culture,2019-06-07,Wikimedia,URLs provided by wikipedia based on the most popular language domains per country
 https://ru.m.wikisource.org/,CULTR,Culture,2019-06-07,Wikimedia,URLs provided by wikipedia based on the most popular language domains per country
+https://github.com/russian-suicide/ololo,PUBH,Public Health,2019-08-06,,GitHub government takedown notice from 2014-10-21
+https://github.com/muromec/blog/blob/d0a4b73064b0d2306a6184b75c9bbfeb3daa2a00/repost/suicide.rst,PUBH,Public Health,2019-08-06,,GitHub government takedown notice from 2014-10-21
+https://github.com/numitus/objidlib/blob/master/suicide.txt,PUBH,Public Health,2019-08-06,,GitHub government takedown notice from 2014-10-21
+https://github.com/samm-git/objidlib/blob/master/suicide.txt,PUBH,Public Health,2019-08-06,,GitHub government takedown notice from 2014-10-21
+https://github.com/l29ah/objidlib/blob/master/suicide.txt,PUBH,Public Health,2019-08-06,,GitHub government takedown notice from 2014-10-21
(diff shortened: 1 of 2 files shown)


muellermartin:
Transparency notice: I've missed a trailing slash for an URL on the Russian list and therefore force pushed to fix the CI error.

lists/ru.csv
```diff
@@ -311,3 +311,34 @@ https://www.svoboda.org/,NEWS,News Media,2018-09-13,OONI,
...
 https://ru.wikisource.org/,CULTR,Culture,2019-06-07,Wikimedia,URLs provided by wikipedia based on the most popular language domains per country
 https://ru.m.wikisource.org/,CULTR,Culture,2019-06-07,Wikimedia,URLs provided by wikipedia based on the most popular language domains per country
+https://github.com/russian-suicide/ololo,PUBH,Public Health,2019-08-06,,GitHub government takedown notice from 2014-10-21
+https://github.com/muromec/blog/blob/d0a4b73064b0d2306a6184b75c9bbfeb3daa2a00/repost/suicide.rst,PUBH,Public Health,2019-08-06,,GitHub government takedown notice from 2014-10-21
+https://github.com/numitus/objidlib/blob/master/suicide.txt,PUBH,Public Health,2019-08-06,,GitHub government takedown notice from 2014-10-21
+https://github.com/samm-git/objidlib/blob/master/suicide.txt,PUBH,Public Health,2019-08-06,,GitHub government takedown notice from 2014-10-21
+https://github.com/l29ah/objidlib/blob/master/suicide.txt,PUBH,Public Health,2019-08-06,,GitHub government takedown notice from 2014-10-21
+https://github.com/kleonov/some-text/,PUBH,Public Health,2019-08-06,,GitHub government takedown notice from 2014-10-21
```
hellais: As these sites are all in HTTPS, to block the sub-page the censor would have to block the entire github.com domain. Maybe we can keep a sample of these that have different domains?

muellermartin: Hm true, I totally missed that. But in case of TLS MITM attacks this might still matter, as only the URLs in question might be blocked while other will pass. Nation-wide TLS MITM is not that far fetched: https://news.ycombinator.com/item?id=20472179

hellais: A TLS MITM would require a custom cert in the user browser. Testing tools should be capable of detecting the TLS MITM before even any request is made.

I see the argument for having one of these to check if they are doing some discrimination between some particular github repos, however I don't think it's necessary to have all of them (test-lists are not block lists).

muellermartin: Also true, but there are governments which are trusted CAs and for them this might be easier then.

But you are right, this isn't a block list. Therefore I will adjust the PR accordingly to only contain one URL of the same (sub)domain each.

PR #504, opened 2019-09-09
Add use-application-dns.net - explicit signal for DNS filtering

R6CG:
This domain is meant to be used as a signal that overt DNS filtering is
in effect. Applications can try to resolve the name as a test for the
presence of DNS filtering. DNS filters that wish to announce their
presence can respond with NXDOMAIN rather than an IP address.

Applications may use the use-application-dns.net to alter their own
behavior. For example, Firefox intends to interpret a
use-application-dns.net NXDOMAIN as a command to disable DNS-over-HTTPS
and use plain old subject-to-tampering DNS instead.

The obvious application of use-application-dns.net, to a network
eavesdropper, is to always return NXDOMAIN, in order to force users onto
insecure DNS. Mozilla says they intend to monitor the use of
use-application-dns.net and stop honoring it if and when ISPs start to
use it.

https://use-application-dns.net/
	This domain is intended to be a content-neutral name that DNS
	filtering products add to their domain block lists.

	This will allow users, as well as browsers and other user agents
	that may implement DNS themselves, to know that the default DNS
	has implemented some sort of content policy that the user may
	wish to maintain, and thus application DNS perhaps should not be
	used, since it would bypass that policy.

	Resolve use-application-dns.net using the platform’s DNS. If the
	result is NXDOMAIN, then there is some sort of content filtering
	or other policy implemented by platform DNS. If the domain
	returns a result, then there is not (or the DNS server has not
	added use-application-dns.net to its filtering lists.)

https://support.mozilla.org/en-US/kb/configuring-networks-disable-dns-over-https
	Network administrators may configure their networks as follows
	to signal that their local DNS resolver implemented special
	features that make the network unsuitable for DoH:

	DNS queries for the A and AAAA records for the domain
	“use-application-dns.net” must respond with NXDOMAIN rather than
	the IP address retrieved from the authoritative nameserver.

https://blog.mozilla.org/futurereleases/2019/09/06/whats-next-in-making-dns-over-https-the-default/
	This canary domain is intended for use in cases where users have
	opted in to parental controls. We plan to revisit the use of
	this heuristic over time, and we will be paying close attention
	to how the canary domain is adopted. If we find that it is being
	abused to disable DoH in situations where users have not
	explicitly opted in, we will revisit our approach.


agrabeli:
LGTM.

PR #472, opened 2019-06-04
add some instagram URLs to ir.csv

xhdix:
To test the hypothesis: "comparison of encrypted content for specific URL censorship in HTTPS"

PR #505, opened 2019-09-11
Remove some URLs related to PR  #472

xhdix:
After checking the links in OONI explorer, I noticed that most of the tests received "URL signature expired" message. And this shows that in Instagram, the URLs of the files change periodically. Therefore, that hypothesis cannot be true for these URLs.

cc: @agrabeli @hellais


agrabeli:
Thanks @xhdix for detecting the issue and opening this PR.

PR #506, opened 2019-09-11
Added VPN to global list

agrabeli:
I've updated the global test list to add a VPN site that's reportedly blocked in Hong Kong (and which is probably worth testing globally). 

See: https://www.privateinternetaccess.com/blog/2019/09/as-protesters-march-on-private-internet-access-vpn-has-been-blocked-in-hong-kong/


agrabeli:
@sneft Merging #514 is a prerequisite to merging this PR.

PR #449, opened 2019-03-23
Updated CM test list

Tafiti:
Added 8 urls from Internet users in Cameroon.

PR #508, opened 2019-09-17
Updated CM test list

agrabeli:
This PR mainly contains an update to the Cameroonian test list based on desktop research. I also added a few URLs to the CU and global test lists (as I came across relevant URLs).


agrabeli:
@sneft I tried resolving the conflict and merging @Tafiti's PR (https://github.com/citizenlab/test-lists/pull/449) but it seems that this PR needs to be merged first.

PR #509, opened 2019-09-19
Resolve CM conflict

agrabeli:
This PR is being opened in an attempt to resolve conflict between #449 and #508.


agrabeli:
I opened this PR to solve issues with #449. This PR includes all the URLs initially added in #449.

PR #510, opened 2019-09-19
Added Aljazeera to global list

agrabeli:
Aljazeera appears to have 2 domains: aljazeera.net & aljazeera.com. 

As part of this PR, I added aljazeera.com to the global test list and fixed the URL for aljazeera.net.

### DIFF
--- a/lists/global.csv
+++ b/lists/global.csv
@@ -350,7 +350,7 @@ http://www.aidsonline.com/,PUBH,Public Health,2014-04-15,citizenlab,Updated by O
 http://www.akdn.org/,ECON,Economics,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 https://www.alarabiya.net/,NEWS,News Media,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
 http://www.aleph.to/,MILX,Terrorism and Militants,2014-04-15,citizenlab,Updated by OONI on 2017-02-14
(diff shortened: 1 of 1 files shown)


agrabeli:
@sneft Merging #511 is a prerequisite to merging this PR.

agrabeli:
Hm, I'm wondering why the Travis CI build is still failing, given that the issues have been fixed with #511 which has been merged....

agrabeli:
@hellais @sneft To delete duplicates (between the global test list and local ones), I opened another PR (#511) which has been merged. I would therefore assume that the checks for this PR should pass, but they don't. How can we make Travis re-process the PR once another PR - which addresses the failed checks - has been merged?

sneft:
I think the issue is that Travis processing only the files contained within your PR, which will technically still contain the duplicate URLs in the local lists since #511 was only merged afterwards. Since there are no conflicts with the base branch, I'm under the impression that if we were to merge this right now it would only add the URLs to the global list and wouldn't change the 3 local lists. 

I think the alternative is for you to update your fork and submit another PR updating this one, which should reflect the changes made to the local lists?

agrabeli:
> Since there are no conflicts with the base branch, I'm under the impression that if we were to merge this right now it would only add the URLs to the global list and wouldn't change the 3 local lists.

@sneft Yes, I have the same impression. Should we try merging this PR?

PR #486, opened 2019-07-29
Add sci-hub.to to German list

muellermartin:
Personal tests and several unrelated sources report that sci-hub.tw is
blocked via DNS in networks of Vodafone at least since December 2017.

Sources:
- https://twitter.com/reinhardremfort/status/938688543572054016
- https://forum.golem.de/kommentare/internet/sci-hub-schwedischer-isp-blockt-elsevier-nach-blockieraufforderung/vodafone-wlan-hotspot-blockt-sci-hub.tw-ebenfalls/121776,5223229,5223229,read.html
- https://www.reddit.com/r/de/comments/8vr8vd/was_zur_h%C3%B6lle_l%C3%A4uft_bei_vodafone_schief/e1pt9yd/
- https://netzpolitik.org/2018/netzsperren-vodafone-muss-kinox-to-blockieren-und-kundendaten-speichern/#comment-2429563
- https://www.heise.de/forum/heise-online/News-Kommentare/Aus-Protest-Schwedischer-Provider-blockiert-Elsevier/In-Deutschland-sperrt-Vodafone-den-Zugriff-auf-Sci-Hub/posting-33375681/show/

PR #518, opened 2019-09-26
Update sci-hub domains in the global list and remove it from DE list

Te-k:
Hi,

Sci-hub is blocked in many countries, and they are updating their domain list regularly (it is easy to follow there https://whereisscihub.now.sh/). So I have removed the domain that are not used anymore (`sci-hub.ac`, `sci-hub.bz`, `sci-hub.org`) and added new one instead (`sci-hub.tw` and `sci-hub.se`).
I noticed that #486 added sci-hub to the German list, if I am right it is not needed as it is in the global list, so I have removed it from the de.csv list.

Let me know if everything is okay.


agrabeli:
LGTM. Thank you @Te-k !

PR #496, opened 2019-08-15
Update tz.csv

Sylviakanari:

PR #520, opened 2019-09-30
Updated TZ test list

agrabeli:
This PR includes:
* Clean up of TZ test list => I manually checked all of the URLs and deleted the ones that were squatted, re-purposed, or otherwise no longer operational.
* Updated URLs
* Fixed category codes
* Added new URLs based on desktop research


anadahz:
@agrabeli Please have a look at https://github.com/citizenlab/test-lists/pull/496 as it contains a lot of information and URLs specific to the Tanzanian test-list.

agrabeli:
@anadahz thanks for the heads-up! I wasn't aware of Sylvia's PR (particularly since it had been merged and the latest version of the TZ test list suggested that it hadn't been updated since 2014). It seems that @hellais has reverted the commit and pushed to master (as you requested), and he's asking that you please re-open a PR with Sylvia's updates.

agrabeli:
@anadahz In my latest commit (https://github.com/citizenlab/test-lists/pull/520/commits/820a9e9ab66670c3863dd3247b4c92584120b961) I have added all of @Sylviakanari 's updates based on PR https://github.com/citizenlab/test-lists/pull/496 and your feedback (i.e. I have addressed your feedback in the commit).

anadahz:
@agrabeli Thank you for working on this PR!

There are still some duplicate domains and other entries that are already included in the global test-list. Please find below the list of duplicates:

* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L14
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L17
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L30
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L32
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L34
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L35
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L36
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L37
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L42
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L43
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L50
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L59
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L66
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L67
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L72
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L75
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L76
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L78
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L87
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L88
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L91
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L99
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L153
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L178
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L186
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L189
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L191
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L220
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L227
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L237
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L243
* https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L256

agrabeli:
@anadahz since this PR passes the Travis CI check, that probably means that some of the duplicates you are referring to are not duplicates in terms of URLs (but duplicates of domains). The Travis script automatically detects the duplicates (between local & global test lists) and flags them (and I have made deletions based on that).

anadahz:
@agrabeli There are quite some duplicates URLs as well as URLs that are already included in the global test-list

For instance the following are the exact same URLs only difference is with `HTTP` or `HTTPS`:
https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L23 and 
https://github.com/citizenlab/test-lists/blob/2a4a68e7743e36ff65fbe5b465ab05f27127b0e9/lists/tz.csv#L237

As a good measure and to save volunteers bandwidth we should ensure that  the test-lists are having the minimum possible URLs/domains except if there is a very specific reason to include same/duplicate entries/domains/URLs. Given that the Tanzanian test-list has 256 entries it may not be a bad idea to remove some (from these same/duplicate) entries, after all most OONI Probe versions are testing both the `HTTP` and the `HTTPS`  (if available) versions of the entries.

agrabeli:
Thank you @anadahz for the feedback. In my latest commit, I deleted URLs based on the duplicates you pointed out (excluding a few URLs that I think are probably worth testing, even though their domains are included in the global list). If you identify any other duplicates, perhaps open a PR proposing their deletion once this PR has been merged (since it seems you can automatically detect them with your script)?

You raise an interesting point. The Travis CI checks are configured to check for duplicate URLs (within the test lists & between the global and country-specific test lists), but not for duplicate domains (presumably your script checks for duplicate domains?). 

On the one hand, we probably don't want to flag all duplicate domains for deletion, as there are cases where we may want to check different HTTP versions of sites (to check the blocking of different web-pages). But on the other hand, we can end up with the problem (as emerged, in fact, in this test list PR) where both the HTTP and HTTPS version of a single domain is added (whether in the same test list, or in both the global and country-specific test list), which is redundant (given that OONI Probe tests for both HTTP and HTTPS anyway, where possible).  

@hellais @sneft  Any thoughts on how to solve this problem? Perhaps we can add new rules to the Travis script?

anadahz:
> Thank you @anadahz for the feedback. In my latest commit, I deleted URLs based on the duplicates you pointed out (excluding a few URLs that I think are probably worth testing, even though their domains are included in the global list). If you identify any other duplicates, perhaps open a PR proposing their deletion once this PR has been merged (since it seems you can automatically detect them with your script)?

Thank you for working on this @agrabeli .

I have listed all the duplicates in https://github.com/citizenlab/test-lists/pull/520#issuecomment-538727846 .
I don't have a script, this is manual process that (usually) ensures that are no mistakes. I manually checked the test-list using the commands `grep`, `cut` and `sort` and then inspecting the differences, then double checked and by using the search function of the browser (shortcut: `ctrl -f`) I could get the line numbers of the test-list in pull requests and pasted them in https://github.com/citizenlab/test-lists/pull/520#issuecomment-538727846

Manually inspection is one of the best processes to eliminate most human errors in test-lists.

> You raise an interesting point. The Travis CI checks are configured to check for duplicate URLs (within the test lists & between the global and country-specific test lists), but not for duplicate domains (presumably your script checks for duplicate domains?).

I guess this may trigger a lot of errors but I may be wrong and a regular expression may find all/most duplicate entries. Neverthelless a good test-list review should happen during the creating, addition and merging of entries in the URL list. Again I don't have any script and I can't think of such a script that may be useful in this case. 

> On the one hand, we probably don't want to flag all duplicate domains for deletion, as there are cases where we may want to check different HTTP versions of sites (to check the blocking of different web-pages). But on the other hand, we can end up with the problem (as emerged, in fact, in this test list PR) where both the HTTP and HTTPS version of a single domain is added (whether in the same test list, or in both the global and country-specific test list), which is redundant (given that OONI Probe tests for both HTTP and HTTPS anyway, where possible).

IMHO I 'll prefer as less same URL/domains entries as possible in the test-lists. Of course this depends on the importance of specific URL entries. Currently the TZ test-list has more than 200 entries, so perhaps it may be more useful to keep only unique domains in order to reduce the size of the test-list.

@agrabeli It seems that you have accidentally committed the UG test-list here. I think it's better if you could add it on a separate pull request.

agrabeli:
@anadahz I have opened a separate PR for the UG test list update (and removed it from here): https://github.com/citizenlab/test-lists/pull/534

hellais:
I rebased and merged this into master.

PR #521, opened 2019-10-02
add new fergana url to RU list

joelanders:
Greetings friends, one of our eQualitie/Deflect clients got blocked in RU yesterday/today (they received advance notice). I went to check the new Explorer (which is great, by the way) and a colleague mentioned that `www.fergananews.com` is an old domain and they've actually switched to `fergana.agency`. So I'm adding the new one here.


agrabeli:
LGTM. Thank you @joelanders !

PR #522, opened 2019-10-05
updat eg.csv list

moeltaher:


agrabeli:
I'm closing this PR because all of the URLs from this PR have already been added through #524 which has already been merged.

PR #528, opened 2019-10-18
Removed blank Nicaragua list (.ni) 

jakubd:
There are no URLs in the Nicaragua list (.ni) list, I believe it should either have a URL added or the list should be removed entirely (this PR).  If this is accepted I can add this logic to the linter as well (that lists should have at least a single URL)

Cheers,
Jakub

### DIFF
--- a/lists/ni.csv
+++ b/lists/ni.csv
@@ -1,1 +1,0 @@
-url,category_code,category_description,date_added,source,notes


agrabeli:
@jakubd good catch! Yep, it would be great if you can add this logic (that lists should have at least a single URL), thank you. Merging this PR in the meanwhile.

hellais:
It appears that deleting this file has broken the logic used by our systems to sync the test list as we were assuming that country lists would only ever be added and never removed.

As a hotfix to unbreak our systems I am going to revert this commit, while we work on coming with some resolution to this problem.

PR #542, opened 2019-11-22
Updated Russia list

sneft:
This update is focused on LGBT content. I've updated a few to reflect URL changes, and added a series of new URLs.

We identified a Ukrainian LGBT organization whose website was on the official RU blacklist, so I've added some of these groups despite this being the Russia list.


sneft:
I inadvertently added a small change to the Canada local list to this PR @agrabeli, but it is relatively minor.

PR #547, opened 2019-12-04
add dns.google

xhdix:


lists/global.csv
```diff
@@ -1300,3 +1300,4 @@ https://equal-eyes.org/,LGBT,LGBT,2019-11-18,OutRight International,
 https://arc-international.net/,LGBT,LGBT,2019-11-18,OutRight International,
 https://www.eurobicon.org/,LGBT,LGBT,2019-11-18,OutRight International,
 https://etherscan.io/,COMM,E-commerce,2019-12-04,OONI,blocked in china according to https://www.coindesk.com/chinas-internet-firewall-has-blocked-access-to-ethereum-block-explorer-etherscan-io
+https://dns.google/dns-query?dns=q80BAAABAAAAAAAAA3d3dwdleGFtcGxlA2NvbQAAAQAB,HACK,Hacking Tools,2019-12-04,Chelchela,
```
hellais: Why is this categorised as hacking?

lists/global.csv
```diff
@@ -1144,7 +1144,7 @@ https://protonmail.com/,HACK,Hacking Tools,2018-03-19,OONI,
 https://getoutline.org/,ANON,Anonymization and circumvention tools,2018-04-06,Duplika,
 http://fteproxy.org/,ANON,Anonymization and circumvention tools,2018-05-21,Community member,
 https://www.thegeekdiary.com/,HACK,Hacking Tools,2018-05-21,Community member,
-https://dns.google.com/,HOST,Hosting and Blogging Platforms,2018-05-02,OONI,
+https://dns.google.com/resolve?name=www.example.com&type=A,HOST,Hosting and Blogging Platforms,2018-05-02,OONI,
```
hellais: I would say that, in light of how the web connectivity test checks for consistency of sites, since you are doing an encrypted HTTPS request we are more likely to accurately detect consistencies/inconsistencies with the index page which returns HTML content vs the JSON payload.

My suggestion is to leave this as is, and if it's useful to have a query, to add it as another row

lists/global.csv
xhdix: I thought it was better for DoH. Which one do you prefer?

HOST = Web hosting services, blogging and other online publishing platforms.
HACK = Sites dedicated to computer security, including news and tools. Includes malicious and non-malicious content.
ANON = Sites that provide tools used for anonymization, circumvention, proxy-services and encryption.

hellais: Now that I think of it, I see what you mean. 

I suppose that for consistency, we should label it like the other DoH service bellow (which are using `HOST`), but we should perhaps extend the definition of the `HOST` category to also include infrastructural services.
cc @sneft

xhdix: done

agrabeli:
LGTM. Thank you @xhdix !

PR #551, opened 2019-12-14
Add Shutterstock URL entry

anadahz:


sneft:
Hi @anadahz,
Good addition, although I think this should be classified as MMED/'Media sharing'

anadahz:
@sneft Thanks for the review.
I changed the category of the entry to `MMED`.

PR #556, opened 2020-01-08
Add Slate to global test list

agrabeli:
Slate is an internationally popular media website, yet it wasn't included in the global test list. It's currently blocked in Pakistan and may be blocked in other countries as well, which is why I opened this PR to add it to the global test list.


agrabeli:
The CI checks are failing because slate.com was also included in the PK test list.

#558 needs to be merged first in order for the CI checks here to pass and this PR to be merged.

PR #561, opened 2020-01-20
update eg.csv

moeltaher:
added 5 LGPT urls


agrabeli:
Thank you @moeltaher for the contributions! I'm closing this PR since it includes the same URLs as #562 (which also includes an extra URL that's not included here).

PR #569, opened 2020-02-03
Updated global test list

agrabeli:
I've added a circumvention tool to the global test list.


agrabeli:
@sneft thanks for the review -- the conflict has been resolved.

PR #576, opened 2020-02-12
Updated Myanmar test list

agrabeli:
This PR includes comprehensive MM test list updates by Kyaw.


sneft:
Tiny thing - row 9 adb.org seems to have an extra / at the end.

agrabeli:
@sneft Good catch! I just fixed that URL.

PR #568, opened 2020-01-31
Update cn.csv

tomac4t:
- Anonymization and circumvention tools `Alexa ranks`
~https://www.uproxy.org/ `#4,919,659`~ (Moved to global list.)
~https://getlantern.org/ `#84,982`~ (Moved to global list.)
~https://getoutline.org/ `#180,667`~ (Duplicated)
- Search Engines
https://searx.me/ `#183,666`
(There is a list about Public Searx instances: https://asciimoo.github.io/searx/user/public_instances.html)
- Social Networking
~https://joinmastodon.org/ `#97,279`~ (Moved to global list.)
https://mastodon.cloud/ `#109,181`
https://mastodon.host/ `#814,718`
~https://mastodon.social/ `#27,011`~ (Moved to global list.)
https://mastodon.xyz/ `#1,350,963`
https://g0v.social/ `#8,057,716`
https://niu.moe/ `#1,190,865`
https://bangdream.space/ 
https://cmx.im/ `#1,668,764`
https://mstdn.io/ `#603,384`
(The federated social network lists: https://the-federation.info/ https://fediverse.network/ etc.)
- News Media
http://www.ftchinese.com/ `#5,380`
http://www.sputniknews.cn/ `#27,471`
~https://www.aljazeera.com/ `#4,114`~ (Duplicated)
https://www.dailysabah.com/ `#24,977`
https://www.dailymail.co.uk/ `#257`
~https://www.theguardian.com/ `#133`~ (Duplicated)
https://www.economist.com/ `#3,106`
https://www.theatlantic.com/ `#1,879`
https://www.nbcnews.com/ `#687`

PR #579, opened 2020-02-16
Update global test list.

tomac4t:
#568


agrabeli:
Thank you @tomac4t , this is great!

Issue #410, opened 2018-11-30
All URLs in the Global List Should be Active URLs.


jakubd:
I realize that there is an intention to keep some URLs that are dead (404ing, parked pages,etc) on the local lists because this repo is intended to measure censorship and this may sometimes last longer than the site itself.  I believe that this logic does not hold as well when talking about the global list.  Is there any objection to checking the consistency of the global list and making sure that all the global list URLs are active non-dead URLs?

sneft:
I agree. Given the frequency with which global list URLs are tested there's an extra incentive to keep it fresh, and knowing the history of that list I'm certain there's some redundant URLs.

At the same time I don't think it needs to be constantly updated - maybe there could be a monthly or quarterly check for dead URLs? It would also probably be useful to have a bit of sanity checking before removing, to prevent removing a URL as a result of a short-term technical issue.

hellais:
This PR has a discussion that is also relevant to this topic I think: https://github.com/citizenlab/test-lists/pull/127

bact:
What about updating HTTP addresses with their HTTPS ones? (if they are automatically redirected)

- This may not applied well for country list, as censorship can go specifically with HTTP address and not HTTPS (and vice versa).
- But for the global list discussed here, I think it could be relevant.

Examples are those Wikipedia links.

PR #580, opened 2020-02-21
Remove http://www.stupidcensorship.com/ (domain parking page).

tomac4t:
Maybe we should remove this domain parking page?

### DIFF
--- a/lists/ae.csv
+++ b/lists/ae.csv
@@ -89,6 +89,7 @@ http://www.dubai-escort-list.com/,PORN,Pornography,2014-04-15,citizenlab,
 http://www.coptic.net/EncyclopediaCoptica/,REL,Religion,2014-04-15,citizenlab,
 http://www.thekoran.com/,REL,Religion,2014-04-15,citizenlab,
 http://www.thequran.com/,REL,Religion,2014-04-15,citizenlab,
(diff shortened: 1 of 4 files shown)


hellais:
Yeah that's a good idea.

Though I do see that it's confirmed blocked in Iran, Saudi Arabia and Arab Emirates: https://explorer.ooni.org/search?until=2020-02-22&domain=www.stupidcensorship.com&test_name=web_connectivity&only=confirmed.

Perhaps we want to move it into those country lists taking note of the fact it's now parked and maybe eventually drop it entirely.

I also wonder if there some scripting we can do to automatically flag parked domains, check if they are known to be blocked in some countries and if so relocate them to the country list flagging them in some way.

For some context around this see: https://github.com/citizenlab/test-lists/issues/410

tomac4t:
I notice that because parking page also cause the false positive (it will redirect to another site in some countries, like blockpage): 
https://explorer.ooni.org/search?until=2020-02-23&domain=www.stupidcensorship.com&probe_cc=CN&test_name=web_connectivity

I move that domain name to ae.csv, ir.csv and sa.csv, and #410 look like long-term job. @hellais

agrabeli:
Thank you @tomac4t for moving this URL to the SA, AE, and IR test lists.

Issue #231, opened 2017-09-29
Add list of URLs to test for censorship in Australia


mozfreddyb:
I'm not sure what the process is for getting something added.
Would I just add the domain to the AU test list?

hellais:
Precisely. In this case we don't have a test list for AU (Australia), so yo would be creating it from scratch.

Use some of the other test lists as a template and be sure to add all the needed column heading (category code, category description, etc.)

We also wrote a more detailed guide on how to contribute to test lists here: https://ooni.torproject.org/get-involved/contribute-test-lists/.

Otherwise if you just comment in this ticket with a list of URLs you would like to add and what their category code is (see: https://github.com/citizenlab/test-lists/blob/master/lists/00-LEGEND-new_category_codes.csv), I can add them for you.

mozfreddyb:
Well, this is the only URL I've got..

hellais:
Well it's a start :) Can you tell me which category it belongs to in the format:

```
http://blog.mish.re/,$CATEGORY
```

and how you would like to be attributed?

mozfreddyb:
It's a pretty harmless security blog, from what I can tell.
I guess the category would be `HACK` then?

Don't know what attribution information you're looking for. 
Name with link? Then [Frederik Braun](https://frederik-braun.com)
You can also omit, I'm not super keen :)

hellais:
> Don't know what attribution information you're looking for. Name with link?
We generally just include some name. It can either be a nick, an organisation name or just community (or anonymous) if you would prefer to not be listed.

mozfreddyb:
anonymous is fine :)
Thank you for preparing the patch in my stead!

hellais:
... and 3 years later I open the pull request: https://github.com/citizenlab/test-lists/pull/581.

Sorry for the delay 😆

PR #581, opened 2020-02-21
Add URL to Australia list

hellais:
Fixes https://github.com/citizenlab/test-lists/issues/231


tomac4t:
The URL no longer exists, the new URL should be added: `https://mishresec.wordpress.com/`.

agrabeli:
@tomac4t You're right. Thanks for sharing the new URL, I just added it to this PR. I've decided to keep the defunct URL because it was previously reportedly blocked, and it could be interesting to check if it's still blocked (even though it no longer exists). Once we've collected some measurements, we could potentially remove it from the list.

PR #583, opened 2020-02-22
Add website with coronavirus dashboard to global

xhdix:


agrabeli:
Thank you @xhdix , LGTM.

PR #584, opened 2020-02-23
Add blocked domains related to pay per view free streaming

fadelkon:
Source: https://cincodias.elpais.com/cincodias/2020/02/18/companias/1582029380_245839.html

Ruling of a court in Madrid, formally against ISP (but claiming party is a subcompany of an ISP), so that all ISP in Spain must block soccer streaming sites and alike.


agrabeli:
Thank you @fadelkon !

PR #589, opened 2020-03-17
Add https://briarproject.org/ to global

bact:
Proposal for the consideration to include https://briarproject.org/ in the global list.

Briar is an app for peer-to-peer encrypted messaging and forums.


bact:
1 errors present

lists/eg.csv (line 873): Duplicate URL between Local List and Global List "https://briarproject.org/"

agrabeli:
Thank you @bact for the contribution! I have removed briar from the EG test list (see https://github.com/citizenlab/test-lists/pull/592) in order to unblock this PR. You're right to point out that this URL should be included in the global test list in order to get tested internationally.

PR #595, opened 2020-03-21
Urgent update to VE list, inc COVID-19 site

andresazp:


agrabeli:
Thank you @andresazp !

PR #597, opened 2020-03-26
Updated MM test list to include porn from blocklist

agrabeli:
This PR includes adult websites that Myanmar ISPs have reportedly been instructed to block access to.


agrabeli:
@sneft I've resolved the conflict.

PR #598, opened 2020-03-27
Add coronavirus sites, bit.ly and discord

anadahz:
Add coronavirus sites, bit.ly and discord that have been reported as
blocked in Portugal.

References:
- https://revolucaodosbytes.pt/a-nos-e-o-bloqueio-de-dominios-legitimos/
- https://revolucaodosbytes.pt/nos-bloqueia-bit-ly/


agrabeli:
Thank you @anadahz ! I have added the HTTPS version of the sites (as they appear when accessed from a normal browser), and I have added the canonical form of the sites, as they appear in a browser (in addition to the domains included in the blocklist).

anadahz:
Thank you @agrabeli .

Please note the git commit https://github.com/citizenlab/test-lists/commit/453e2f96d22e760e9943d1cbd88ddb9cfec3c632 has 2 entries of `bit.ly`.

agrabeli:
@anadahz yes, that's done intentionally. Normally we add URLs exactly as they appear in a browser, so I was initially planning on replacing bit.ly with bitly.com (since that's what it redirects to). However, since the blocklist mentions "bit.ly", I figured it's probably worth including both versions, in case they're blocking the one, but not the other (or both). For the same reason, I've also added 2 versions of the discord app.

PR #602, opened 2020-04-02
[ES] Add a torrent sharing website

fadelkon:
Blocked at least at Vodafone, usign Allot it's become usual. "Por causas ajenas a Vodafone, esta web no está disponible"


agrabeli:
LGTM. Thank you @fadelkon !

PR #607, opened 2020-04-07
fixed future/impossible date_added from uy list

jakubd:
I had some analysis break because date_added was a date in the future (year 3027) in one case, this fixes that (based on surrounding date)  I guess the linter should make sure this doesn't happen but not a big deal.


hellais:
🐳

PR #606, opened 2020-04-07
Removed duplicates to appease lint_lists on master

jakubd:
lint_lists output was complaining about duplicates on master this PR addresses.

Output goes from:

```
Analyzed 32902 URLs in 142 countries
9 errors present
lists/ua.csv (line 607): Duplicate URL between Local List and Global List "https://sputniknews.com/"
lists/my.csv (line 97): Duplicate URL between Local List and Global List "https://medium.com/"
lists/eg.csv (line 398): Duplicate URL between Local List and Global List "https://sputniknews.com/"
lists/eg.csv (line 608): Duplicate URL between Local List and Global List "https://medium.com/"
lists/eg.csv (line 872): Duplicate URL between Local List and Global List "https://about.riot.im/"
lists/sy.csv (line 251): Duplicate URL between Local List and Global List "https://livestream.com/"
lists/sy.csv (line 333): Duplicate URL between Local List and Global List "https://sputniknews.com/"
lists/global.csv (line 1362): Duplicate URL "https://www.who.int/"
lists/global.csv (line 1368): Duplicate URL "https://yandex.com/"
```

to 

```
Analyzed 32893 URLs in 142 countries
ALL OK
```

Issue #604, opened 2020-04-04
Switch to Github actions?


tomac4t:
Looks something wrong with Travis CI. Recently, I notice that it doesn't build for Pull Request (in this repo). Maybe should switch to Github actions? Like: 
```yaml
name: check
on: 
  push:
    paths:
    - "lists/*"
  pull_request:
    paths:
    - "lists/*"
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2
    - name: check
      run: python scripts/lint-lists.py lists/
```

jakubd:
Agreed, tested this on a toy copy of the repo and works well.  I am not sure what the issue is with Travis all the builds are done on the Travis side but the results don't always get posted.

Added your script as the basis in a PR for now. see #606 for more discussion

tomac4t:
Move `.github/main.yml` to `.github/workflows/main.yml`? I believe the `.yaml` file only work on `.github/workflows` directory. @jakubd 

Reference: https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#creating-a-workflow-file

jakubd:
You got it!  Fixed some of the names to be a bit more descriptive as well. Thanks!

jakubd:
This is merged in now, so no more Travis.  Thanks for the help!

PR #608, opened 2020-04-08
Add github actions to run lint_lists

jakubd:
This adds a Github Action that runs the lint_lists script against all lists on push/pulls.

see #604 , and #606


jakubd:
as requested, I rebased and dealt with the backlog of lint_lists errors here (this should includes changes that makes #606 irrelevant)

hellais:
Is the action running? I still only see the output of the travis check and not the github action.

Screenshot 2020-04-13 at 17 45 34

When going into the "Actions" tab there doesn't appear to be any action setup:

Screenshot 2020-04-13 at 17 46 37

Maybe something needs to be done to enable the action on the repo? Or could it be that it doesn't work because you are pushing to a fork of it instead of the upstream?

PR #611, opened 2020-04-11
Update dz.csv

slindy:
Proposal for adding banned domain.


BoFFire:
Please see : https://github.com/citizenlab/test-lists/pull/612

Issue #613, opened 2020-04-13
Add occrp.org to global list in category "news"


DrWhax:
Hi!

We'd love to have occrp.org added to the global list testing list. We cover most of eurasia, central asia, parts of africa and latin-america. In some of these countries we keep hearing reports we're blocked and we'd like to have better insight in whether that's the case.

For example, we've gotten recently blocked in venezuela because of publishing about a huge corruption scandal involving venezuelan generals.

Thanks for your consideration! :)

DrWhax:
Seems we're done here! :)

PR #614, opened 2020-04-14
add occrp.org to global csv and remove from az list

DrWhax:
In response to issue #613


hellais:
This is great. Thanks!

Merging.

PR #620, opened 2020-04-25
Update dz.csv

BoFFire:
Adding dzvid.com website reportedly blocked for internet users in Algeria.
Source article : https://www.dzvid.com/2020/04/24/le-site-dzvid-censure-en-algerie/


agrabeli:
Thank you @BoFFire !

PR #622, opened 2020-05-06
add ocsp.apple.com

xhdix:
Used for the certificate validation in iOS, tvOS, and macOS
https://support.apple.com/en-us/HT210060

https://twitter.com/search?q=ocsp.apple.com%20127.0.0.1&src=typed_query
It is not yet clear why users do this. :man_shrugging:

### DIFF
--- a/lists/global.csv
+++ b/lists/global.csv
@@ -1401,3 +1401,4 @@ https://www.freepik.com/,MMED,Media sharing,2020-04-01,OONI,URL shared by commun
 https://www.f-list.net/,GAME,Gaming,2020-04-01,OONI,URL shared by community member
 https://www.1800respect.org.au/languages/,XED,Sex Education,2020-04-01,OONI,URL shared by community member
 https://www.sbs.com.au/language/coronavirus?cid=infocus,NEWS,News Media,2020-04-01,OONI,URL shared by community member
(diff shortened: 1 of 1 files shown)


smege1001:
They do this to prevent app revokes btw

hellais:
What is the base64 encoded path in the URL? is it a special value that is likely to change or is it the OCSP staple for a particular resource?

Do we need to set it to a specific value or can we use a generic landing page for it to check if it's blocked?

xhdix:
@hellais, I got it from here:
https://github.com/elastic/beats/blob/master/x-pack/filebeat/module/zeek/http/test/http-json.log
And you can see many of these here:
https://urlscan.io/search/#page.domain%3Aocsp.apple.com
Like:
 http://ocsp.apple.com/ocsp04-applerootcag3/ME4wTKADAgEAMEUwQzBBMAkGBSsOAwIaBQAEFDQ9%2BWuiBh%2FOOBXswAMX5FRQRzCbBBS7sN6hWDOImqSKmd6%2Bveuv2sskqwIISW0vvzqY2pc%3D 
I'm not sure, but I don't think it will change.

However, if receiving `<html><head><title>Error</title></head><body>403 - Forbidden</body></html>` is fine, I can change it.

xhdix:
It doesn't seem necessary.

PR #624, opened 2020-05-17
add lakome in ma.csv

smege1001:
In Maroc Telecom, the connection gets reset
In Orange, redirects to google
haven't tested inwi


hellais:
We prefer to add URLs to the lists that are still active. If they are already in the list and are blocked, then we may keep them, but I think it's preferable to not add sites that are already down.

Would also like input on this from @agrabeli and @sneft

agrabeli:
@smege1001 I generally agree with the point that @hellais made. When and why was this news website blocked? Can you please share more information on the importance/relevance to continue testing it? Thank you.

PR #625, opened 2020-05-18
Added 3 News Media sources to br.csv

nunesgh:
Hi,

It is my first contribution to Citizen Lab's test lists. I have added three News Media sources to the Brazilian file, namely:

- [economia.uol.com.br](https://economia.uol.com.br)
- [noticias.uol.com.br](https://noticias.uol.com.br)
- [theintercept.com/brasil/](https://theintercept.com/brasil/)

The first two are the economy-related website and the general-news website from major local UOL portal. The third one is the pt-br branch of The Intercept, run in Brazil by Glenn Greenwald.

Based on the news coverage by those websites and recent statements from Brazil's president and members of his family, I think those three websites should be added to the br.csv file.

Thank you.

### DIFF
--- a/lists/br.csv
+++ b/lists/br.csv
@@ -852,9 +852,7 @@ http://terceirotempo.bol.uol.com.br/,CULTR,Culture,2016-06-08,OONI,Blog about sp
 http://terezacs.blogspot.com/,POLR,Political Criticism,2017-08-28,Artigo 19 + CryptoRave + Coding Rights,Blogs de direita (Mapeamento Artigo 19)
 http://terezopolis.go.gov.br/,GOVT,Government,2017-08-30,Artigo 19 + CryptoRave + Coding Rights,"Portais dos municípios analisados pela Escala Brasil Transparente (1ª e 2ª edição), http://relatorios.cgu.gov.br/Visualizador.aspx?id_relatorio=1802"
 http://thaumaturgo.ac.gov.br/site01/index.php,GOVT,Government,2017-08-30,Artigo 19 + CryptoRave + Coding Rights,"Portais dos municípios analisados pela Escala Brasil Transparente (1ª e 2ª edição), http://relatorios.cgu.gov.br/Visualizador.aspx?id_relatorio=1196"
-https://theintercept.com/,NEWS,News Media,2017-08-28,Artigo 19 + CryptoRave + Coding Rights,Jornalismo independente (Mapeamento Artigo 19)
 https://theintercept.com/2016/05/23/new-political-earthquake-in-brazil-is-it-now-time-for-media-outlets-to-call-this-a-coup/,POLR,Political Criticism,2016-06-10,OONI,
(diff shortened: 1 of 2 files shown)


lists/br.csv
```diff
@@ -852,6 +854,7 @@ http://terezopolis.go.gov.br/,GOVT,Government,2017-08-30,Artigo 19 + CryptoRave
 http://thaumaturgo.ac.gov.br/site01/index.php,GOVT,Government,2017-08-30,Artigo 19 + CryptoRave + Coding Rights,"Portais dos municípios analisados pela Escala Brasil Transparente (1ª e 2ª edição), http://relatorios.cgu.gov.br/Visualizador.aspx?id_relatorio=1196"
 https://theintercept.com/,NEWS,News Media,2017-08-28,Artigo 19 + CryptoRave + Coding Rights,Jornalismo independente (Mapeamento Artigo 19)
 https://theintercept.com/2016/05/23/new-political-earthquake-in-brazil-is-it-now-time-for-media-outlets-to-call-this-a-coup/,POLR,Political Criticism,2016-06-10,OONI,
+https://theintercept.com/brasil/,NEWS,News Media,2020-05-17,nunesgh,
```
hellais, line 857: I think we should be adding this to the global list, without the `/brasil` path.

Since the site is HTTPS if we test the homepage it's the same as testing the subdomains.

nunesgh: Agreed! I did not check whether theintercept.com was already on the global list or not before adding the specific path for Brazil here. I will pay attention to that in the future. Thank you for your feedback, @hellais!

But what is the difference when it comes to adding a HTTPS instead of a HTTP website? How does it change how the website is tested?

Also, I have noticed several specific URLs in the br.csv file related to specific news articles from some News Media websites. At least some of those URLs are outdated in the sense that they cover events from years ago. Should they be kept there, or would it be the case to remove them while making sure the respective News Media websites are covered by the list?

For example, both noticias.terra.com.br and noticias.uol.com.br were not on the list, but specific news articles from those portals were. In both cases, news articles from 2016.

agrabeli: Thank you @nunesgh for your contributions! 

You're right to point out that outdated news articles or websites that are otherwise no longer relevant should be removed from the test list. We'd greatly appreciate any further suggestions you may have in terms of other news websites to add to the Brazilian test list, or outdated/irrelevant URLs that should be removed. We share further information here: https://ooni.org/get-involved/contribute-test-lists

Regarding the difference between HTTP or HTTPS sites: If a site is hosted on HTTPS, a censor cannot limit the blocking to a specific webpage (for example: https://theintercept.com/2020/05/25/coronavirus-tracking-bracelets-monitors-surveillance-supercom/), but they would have to block access to the entire website (for example: https://theintercept.com/). Therefore, if a site supports HTTPS, we only add the canonical HTTPS version of that website to a test list.

PR #626, opened 2020-05-18
add project Gutenberg

eshellman:
Public domain site targeted by copyright extremists


hellais:
We generally prefer to have added to the test lists the canonical URL of the site. In this case I see that `https://gutenberg.org/` does a 301 redirect to http://www.gutenberg.org/, so I think that is the URL we should be adding to the test list.

If you change the URL to include the `www.` I would say this is good to merge.

PR #628, opened 2020-05-20
Added Let's Encrypt to global test list

agrabeli:
This PR adds letsencrypt.org to the global test list, as well http://ocsp.int-x3.letsencrypt.org/ which is reportedly blocked in China, but probably worth testing globally.


agrabeli:
There's something wrong with the CI checks.

hellais:
Yeah it might be a problem with github or some config issue.

@jakubd can you take a look at the github actions setup of citizenlab?

hellais:
I am merging this as it was suggested by a community member and I think we can be the reviewers in such case.

PR #631, opened 2020-06-04
add Discord main domain.

xhdix:


agrabeli:
Oh, I'm surprised discord wasn't already included in the global test list. Thank you @xhdix for adding it!

PR #635, opened 2020-06-13
move tanzil.net to global list and add quran.com

xhdix:
https://www.uptrends.com/tools/uptime

image

image


agrabeli:
Thank you @xhdix !

PR #636, opened 2020-06-19
added global pride site to global list

jakubd:
Added URL for the global pride event to the global testing list.


agrabeli:
Thank you @jakubd !

PR #639, opened 2020-07-16
add trello.com and some DoH servers to global

xhdix:


agrabeli:
LGTM. Thank you @xhdix !

PR #654, opened 2020-07-28
Updated list with few records

ifreedombelarus:


agrabeli:
Thank you @ifreedombelarus !

PR #661, opened 2020-08-21
Updated several lists with reproduction rights sites

agrabeli:
These are sites that host information similar to womenonweb.org and may therefore be blocked in the countries whose test lists they've been added to (as they block access to womenonweb.org).


hellais:
@agrabeli this needs to resolve some conflicts

agrabeli:
@hellais @sneft I have resolved all conflicts (and deleted duplicates). I think this PR may be ready for merging now.

PR #507, opened 2019-09-13
add full blacklist of blocked gambling sites in switzerland (Comlot and ESBK)

mritzmann:
Access to these websites has been blocked in accord-ance with Art. 86 et seq. of the Federal Act on Gambling and Art. 88 et seq. of the Gambling Ordinance.

source: https://blacklist.comlot.ch/comlot_blacklist_20190903.pdf

PR #670, opened 2020-09-25
update blacklist of blocked gambling sites in switzerland

mritzmann:
All these websites must be blocked by swiss internet providers using dns blocking. Same as in #507. 

Source:
* https://blacklist.comlot.ch/comlot_blacklist_20200818.txt 
* https://www.comlot.ch/en/gambling-in-switzerland/access-blocking/blacklist


hellais:
I suspect we already have enough gambling websites in the test-list. @agrabeli and @jakubd can also weigh in, but I think we might not want to merge this.

@mritzmann see: https://ooni.org/support/faq/#what-is-the-difference-between-a-test-list-and-a-blocklist

mritzmann:
Guess you're right.

PR #673, opened 2020-09-30
improve docs 

kartikey6555:
it can be written as more than one or more country


hellais:
Thanks for this! Merging.

PR #679, opened 2020-10-19
Add URLs related to the demonstrations in July-October 2020

bact:
Thai government now consider using Emergency Decree to block some websites and Telegram
https://prachatai.com/journal/2020/10/90030

- Add variations of change.org that got blocked in Thailand (https://, http://, with and without www. got different blocking results)
- Add URLs news websites mentioned in the government order
- Add URLs of three demonstration organizers and one rescue group


hellais:
LGTM!

PR #682, opened 2020-11-02
Updated TZ test list

agrabeli:
These updates include URLs shared by community members (and their testing presents signs of blocking in TZ).


jakubd (approved):
Looks good to me.

PR #692, opened 2020-11-30
Drop dead URL from global list

hellais:


Sami32:
Broken website detected as A"nomaly":
IMG_20201130_154305

PR #706, opened 2021-01-04
Adding potentially blocked china covid urls

jakubd:
I am proxying this update for Phong (OTF) but his work seems to indicate these six domains are being DNS poisioned within China.  Hoping to see the breadth of this behaviour with this update.

Cheers,
Jakub


hellais:
LGTM

PR #710, opened 2021-01-07
added hk chronicles

jakubd:
Added HK site reportedly blocked according to 

* https://twitter.com/KwongGlacier/status/1347176882664443909
* https://twitter.com/lokmantsui/status/1347196360735346691

Probably worth having on list.


hellais:
LGTM!

PR #711, opened 2021-01-07
Updated HK test list

agrabeli:
Adding https://hkleaks.info/ which is reportedly blocked in HK too.


jakubd:
Nice catch! Approved and merging.

PR #713, opened 2021-01-11
Add app store links to Global test list

agrabeli:
The Play Store and App Store are reportedly blocked in Uganda. I figured it's probably worth having them tested globally, which is why I added relevant links to the global test list.


jakubd:
The substance of the update looks good to me except the merge conflict.

PR #716, opened 2021-01-12
Proposed updates to Uganda test list to include regional and global news sites

Maundu333:
I propose changes to the test list for Uganda, to enable us to collect data on specific action that might be taken against news sites. Most local news sites were already reflected on the list, but not regional or international outlets.


hellais:
Thanks for the contribution! These URLs need to be formatted properly in order for us to merge them.

agrabeli:
Thank you @Maundu333 for the great contributions! I fixed the formatting and I removed URLs which are already included in the Global test list (and therefore tested globally).

PR #717, opened 2021-01-13
Update to jo.csv

rayasharbain:
Added VPN provider websites: TunnelBear, NordVPN, Psiphon (x2). Noticing that some are blocked on some local ISPs.


hellais:
I think these should be added to the global list as they are of global relevance.

agrabeli:
The 3 URLs (added in this PR) are already included in the Global test list, meaning that they are by default tested in Jordan and every other country in the world. I am therefore closing this pull request. 

@rayasharbain thanks anyway! We hope to see more contributions from you. <3

PR #718, opened 2021-01-15
Patch 3

oussjarrousse:


hellais:
Perhaps this should be included in the global test list as it's of global relevance

PR #720, opened 2021-01-19
Update Hong Kong list to account for developments.

nathanhammond:
- hkleaks.info and hkchronicles.com are both the same site. They're closer to political criticism than they are to a news source.
- Four days ago the administrator of the site created a *third* domain pointing to the same site, blockedbyhk.com.
- The first two domains were both included in the first-known instance of using the NSL to block websites in Hong Kong.
- The third domain, blockedbyhk.com, as of the date of this PR remains unblocked.


nathanhammond:
For additional context you can start with this thread on Twitter where I have been tracking it: https://twitter.com/nathanhammond/status/1351376991740710915

PR #722, opened 2021-02-04
Add Namecoin.org

JeremyRand:
Seems on-topic given that other censorship-resistant suTLD's such as Tor, DNS providers such as OpenDNS, and other blockchain projects such as Bitcoin are on the list.  Also seems sufficiently non-redundant given that I don't see any other decentralized DNS-like naming systems on the list.  I got a user report alleging that it was blocked in Belarus in January 2021 (I haven't been able to verify the accuracy of that report).

I initially reached out on IRC; @agrabeli recommended that I open a PR here.


agrabeli:
Thank you @JeremyRand!

PR #724, opened 2021-02-09
Add Démocratie Participative

hanscalder:
This site is alleged to be blocked. It may be a good idea to add it to other Francophone lists as well.


hellais:
LGTM

PR #726, opened 2021-02-09
Add Library Genesis URLs

hanscalder:
Add Library Genesis URLs. Libgen infringes DMCA, blocked in countries all over the world. Certain mirrors respect DMCA, should be blocked less.

Forums also added, though should be less liable for blocking.

Add all mirrors that were up at the time of checking (9/2/21), according to precedent set by sci hub.


agrabeli:
I have removed dailymotion from the SY list (https://github.com/citizenlab/test-lists/pull/884) in order to pass the CI checks for merging this PR (since dailymotion is already included in the Global test list).

agrabeli:
Thank you @hanscalder

PR #727, opened 2021-02-10
Prune dead urls

hellais:
This is a big pull request which automates the process of deleting all dead URLs from the testing lists.

I wrote a script that looks for expired domains and drops them from the relevant testing list.

All in all, it leads to 2274 URLs being deleted.


jakubd (commented):
Looks great to me, and a good idea at this point.  I spot checked some of the removals and read through the URLs and can't find any objection here.  Thanks for taking this on!

One small point to clarify: are these removals based on a single pass through the list?  If so I would recommend doing it a few times to rule out any momentary DNS failures.  Other than that this looks fantastic.

hellais:
Thanks for the review @jakubd!

> One small point to clarify: are these removals based on a single pass through the list? If so I would recommend doing it a few times to rule out any momentary DNS failures. Other than that this looks fantastic.

Yes this makes sense. I will see to add support for this in the script and make another double pass at pruning the test list of dead URLs updating the PR accordingly.

hellais:
@jakubd I updated the script to do two passes at resolutions, the second pass randomises the order in which the domains are looked up. I also reduced the level of parallelism so that we can further reduce the likelyhood of transient failures.

Does this looks good to you?

jakubd:
Looks good to me (and to @sneft), go ahead with the merge (this is a big one so I will leave to you to merge)

Good work!

PR #729, opened 2021-02-22
Add tests for Signal Messenger

arky:
Added tests for Signal messenger as stop-gap measure until the [signal messenger tests](https://github.com/ooni/probe-cli/pull/230/files#diff-ed1b4f0013a12402e0f5ced63481629b1606b49c75d2e4a91d7081df31ebafa4R112) are implemented in the new release.


agrabeli:
Thanks @arky ! Normally such URLs would go under the Global test list (https://github.com/citizenlab/test-lists/blob/master/lists/global.csv), but given the urgency, I think it's OK to have them temporarily in the MM test list.

arky:
Thank you @agrabeli Noted, I didn't know about the Global Test list. 

 I think there are more comprehensive support for signal is coming up in this PR https://github.com/ooni/probe-cli/pull/230/files#diff-ed1b4f0013a12402e0f5ced63481629b1606b49c75d2e4a91d7081df31ebafa4R112

PR #730, opened 2021-02-22
Update lists/global.csv to add a Wikimedia project

azadi:
Add upload.wikimedia.org that hosts media files for the Wikimedia Foundation to the list.


hellais:
LGTM!

PR #738, opened 2021-03-05
Another blocked resource in Belarus

abitrolly:
The specific page why most likely these stats about people beaten by police

https://mediazona.by/article/2020/10/13/minsk-beaten


hellais:
LGTM

PR #742, opened 2021-03-12
Explain why URL is invalid when lint fails

abitrolly:
This also makes `lint-lists.py` executable.


hellais (approved):
Thanks for the contribution! LGTM

PR #743, opened 2021-03-15
Add provider to global list

frozen-sky:
Add Greenhost main site, a provider focussing on digital rights and privacy to the global list


hellais:
LGTM

PR #750, opened 2021-03-31
Added some missing mm gov sites

jakubd:
Added some URLs of MM government sites that were missing from the local list.


hellais:
LGTM

PR #751, opened 2021-04-01
Add ipfs.io to BE list

anadahz:
ipfs.io reported as blocked in Belgium
https://web.archive.org/web/20210401073940/https://discuss.ipfs.io/t/ipfs-blocked-in-belgium/7930


hellais:
I agree with the URL suggestion, but I think it's miscategorised. I believe it should be categorised as:
```
File-sharing | FILE
```

anadahz:
@hellais thanks for reviewing the PR.
I fixed the category name in: https://github.com/citizenlab/test-lists/pull/751/commits/87cbe1a882340fb720229ca9c9e2f498aad76875

hellais:
LGTM

PR #753, opened 2021-04-06
Remove www.sexualhealth.com

mimi89999:
Probably the domain has a new owner now. It seems that for a year it was entirely down and now it's either redirecting to a porn site or entirely inaccessible:
https://web.archive.org/web/20210901000000*/www.sexualhealth.com
https://explorer.ooni.org/search?until=2021-04-07&since=2021-03-07&domain=www.sexualhealth.com


hellais:
LGTM

PR #754, opened 2021-04-06
Add CTRL for PL CERT lista ostrzezen

mimi89999:
Those seem to be control domains used by them. Discovered through own analysis of the list.


jakubd:
Looks good to me, and this is a very economic way of testing this particular filtering policy with not a lot of URLs added to this list.

jakubd (approved):

PR #756, opened 2021-04-14
Remove www.connotea.org

mimi89999:
Website is down. Was last seen in 2016: https://web.archive.org/web/20160830114545/http://www.connotea.org/
> Connotea discontinued service on March 12, 2013.


hellais:
LGTM

PR #758, opened 2021-04-14
Remove www.radioislam.org

mimi89999:
Down since https://web.archive.org/web/20180324210936/http://radioislam.org/


hellais:
I am not sure we should be removing this URL as we still see it being blocked in places such as Russia: https://explorer.ooni.org/measurement/20210416T061324Z_webconnectivity_RU_41661_n1_PMsjWtL6gxGVwcwq?input=http%3A%2F%2Fwww.radioislam.org%2F.

Maybe we want to move it to the country specific list, but drop it from the global list?

mimi89999:
I disagree. This URL might be removed from the block list in future (but some block lists probably only remove content on request). The worst issue is like in https://github.com/citizenlab/test-lists/pull/753 that the domain owner might change and with that the content of the page. In that case, the site can be blocked or unblocked for different reasons, but for an observer that did not check carefully it could seem that there was a change in the policy.

agrabeli:
I agree with @mimi89999. I think we shouldn't be including too many non-operational websites in the Global test list (and only include them in "special" cases). I think the Global test list deserves a clean-up of non-operational sites (and I frequently see 'radioislam.org` leading to false positives in measurements).

PR #759, opened 2021-04-14
Remove international.ohmynews.com

mimi89999:
Last content on the website was from 2013, so it's not surprising that it's gone: https://web.archive.org/web/20200721180116/http://international.ohmynews.com/


hellais:
There is one recent measurement from Italy that shows it being blocked (https://explorer.ooni.org/measurement/20210412T103824Z_webconnectivity_IT_30722_n1_PVPXqsKhpyYYofoo?input=http%3A%2F%2Finternational.ohmynews.com%2F), but it appears to be a local network filtering device rather than country level filtering, so I think it's OK to drop it.

PR #760, opened 2021-04-14
Remove www.arabhra.org

mimi89999:
Down since 2019: https://web.archive.org/web/20190102003828/http://www.arabhra.org/


hellais:
LGTM

PR #762, opened 2021-04-14
Remove www.efindlove.com

mimi89999:
Down since 2017: https://web.archive.org/web/20170916084953/http://efindlove.com/. Now parked.


hellais:
LGTM

PR #765, opened 2021-04-21
Remove www.crackz.ws

mimi89999:
Parked at least since 2019: https://web.archive.org/web/20190114201444/http://www.crackz.ws/


hellais:
LGTM

PR #767, opened 2021-04-27
[th] Remove trailing slash (/) from links to Wikipedia articles

bact:
URL like `https://xx.wikipedia.org/wiki/xxx/`, with a slash (`/`) at the end, will return an article not found notice from Wikipedia.

All Wikipedia article URLs should not have a trailing slash.

Change all `https://xx.wikipedia.org/wiki/xxx/` to `https://xx.wikipedia.org/wiki/xxx`


hellais:
LGTM

PR #770, opened 2021-05-16
Updated the IL test list

agrabeli:
In light of the situation in Palestine, I think it would be interesting to check whether Palestinian news websites and human rights sites are being censored in Israel. I have therefore updated the IL test list accordingly (while also adding a few Israeli news websites). 

As this seems high priority for testing, it would be great if we could merge this PR asap.


hellais:
LGTM

PR #776, opened 2021-05-31
Update ua.csv

readonly647:
Added blocked russian media and gambling websites


hellais:
There is an issue in the formatting of the date in the submitted list. We also generally try to avoid submitting full blocklists. I would suggest only including a subset of all these URLs, 5-6 of the most popular gambling sites.

mScott224 (approved):

PR #792, opened 2021-07-06
Update jo.csv

rayasharbain:
Added clubhouse app homepage URLs


rayasharbain:
It seems that one of the URLs is already on the global list, will add the second to the global list as well and close this pull request.

PR #806, opened 2021-07-30
added current CUII recommendations to global list

mhombach:
The official German company, who is providing "transparent censorship", is the CUII (https://cuii.info).
On the subpage (https://cuii.info/empfehlungen/) you can see a list of which pages are officially recommended to be censored in Germany (by DNS) and German ISPs have to follow those recommendations.
There is no judge/law involved here, it's just private music/movie lobby paying money to certain people.

This list is currently growing very slowly because CUII wants to fly under the radar, but I guess the list will grow faster over time.
If this is ok with citizenlab, I want to regularly check on the CUII and update the global list.

Also: I chose the `FILE` category for this, because those pages are basically dictionaries for File-Hosters and I didn't want them to be mixed up with websites really writing about media/music/movies/games. If that's not a good choice, please tell me which category those sites should be going to.


lists/global.csv
```diff
@@ -1465,3 +1465,7 @@ https://clubhouse.pubnub.com/,GRP,Social Networking,2021-04-04,Chelchela,
 https://brave.com/,HACK,Hacking Tools,2021-06-24,Chelchela,
 https://search.brave.com/,SRCH,Search Engines,2021-06-24,Chelchela,
 https://startpage.com/,SRCH,Search Engines,2021-06-24,Chelchela,
+https://s.to/,FILE,CUII-Recommendation (cuii.info),2021-07-30,mhombach
+http://canna.to/,FILE,CUII-Recommendation (cuii.info),2021-07-30,mhombach
+https://nsw2u.com/,FILE,CUII-Recommendation (cuii.info),2021-07-30,mhombach
+https://newalbumreleases.net/,FILE,CUII-Recommendation (cuii.info),2021-07-30,mhombach
```
hellais, line 1471: There is an issue with the format of these URLs in the test list. The third column should be the description of the category code, in this case it should read `File-sharing` instead of `CUII-Recommendation (cuii.info)`

mhombach: > 

Hey, sorry I didn't get the notification for your reply here for some reason.
I can totally change that, my idea was just, that those links would be blocked by a specific company/organization and so we could group them together... Do you think there some way lateron to implement some "notes/comments" into a list entry, which could be displayed in the app?
But for this PR I can just change that column, thanks for the response and feedback!

mhombach: @hellais Fixed everything, can you have a look again? =)

hellais:
LGTM

PR #807, opened 2021-07-30
Remove defunct askmoses.com

evilaliv3:
Reference: https://en.wikipedia.org/wiki/AskMoses.com

Fixes OpenObservatory/ooni-resources#22, fixes ooni/probe#1712


hellais:
LGTM

PR #808, opened 2021-08-03
feat(global.csv): add possibly-geoblocked-outside-china URLs

bassosimone:


bassosimone:
@sneft do you perhaps have time to look at this? thank you! 🙏

hellais:
We are going to land this although the sites are not particularly popular, but it's useful for doing an investigation on server-side blocking from these sites.
We are going to keep it in master for a few months and then drop them.

PR #809, opened 2021-08-06
Update bd.csv

Oymate:
It would be better if you committed this yourself(maintainer)


lists/bd.csv
```diff
@@ -277,3 +277,11 @@ http://gazwah.net/,NEWS,News Media,2018-12-13,citizenlab,Reported to be blocked
...
 https://lookeast.in/,NEWS,News Media,2018-12-13,OONI,Reported to be blocked
 https://womenchapter.com/,HUMR,Human Rights Issues,2018-12-13,OONI,Reported to be blocked
+https://vk.com
+https://medium.com
+https://bit.ly
+https://archive.org
+https://issuu.com
+https://mediafire.com
+https://change.org
+https://lookeast.in
```
hellais, line 287: These URLs are missing the category_code, category_description, date, author and notes columns

PR #826, opened 2021-09-25
Add domain names for iCloud Private Relay

R6CG:
iCloud Private Relay is an onion routing proxy available in some versions of iOS, iPadOS, and macOS. Apple's advice to network operators who want to block access to Private Relay is to configure the DNS server to respond negatively for queries for the two names added in this commit, mask.icloud.com and mask-h2.icloud.com.

["Allow for network traffic audits"](https://developer.apple.com/support/prepare-your-network-for-icloud-private-relay)
> The fastest and most reliable way to alert users is to return a negative answer from your network’s DNS resolver, preventing DNS resolution for the following hostnames used by Private Relay traffic.

Some caveats to adding these domain names to test-lists:
* We really only care to test a DNS resolution for these names, not actually attempt to fetch a web page. A similar situation applied with #504.
* The documented protocol of Private Relay is anyway UDP+QUIC (HTTP/3), not TCP+TLS. I am guessing, purely based on the name, that mask-h2.icloud.com may be an HTTP/2-based fallback for mask.icloud.com.
* Apple clients [reportedly](https://github.com/net4people/bbs/issues/87) send not only A queries for these names but also [HTTPS (RR type 65)](https://datatracker.ietf.org/doc/html/draft-ietf-dnsop-svcb-https-07#section-14.2) queries.

Being able to resolve mask.icloud.com and mask-h2.icloud.com does not guarantee that Private Relay will actually work. Apple self-censors the service in certain geographic regions. I don't know how Apple's own restrictions are enforced.

Various other domain names may be involved in a Private Relay
connection. This commit includes only the ones that are documented by
Apple to use for blocking access.
* mask.apple-dns.net: mask.icloud.com is a CNAME for this name.
* mask-t.apple-dns.net: mask-h2.icloud.com is a CNAME for this name.
* mask-api.icloud.com: Apple clients were seen to make queries for this name in experiments.
* mask-api.fe.apple-dns.net: mask-api.icloud.com is a CNAME for this name.


witchof0x20:
Seconding this

https://9to5mac.com/2022/01/10/t-mobile-block-icloud-private-relay/

There are (unverified) reports of potential carrier censorship of Apple private relay, and it would be useful to be able to measure this.

hellais:
lgtm

PR #846, opened 2021-10-07
Removed http://www.roulette.sh/

RandomGuy48:
Parked domain


agrabeli:
Thank you @RandomGuy48

PR #847, opened 2021-10-07
Removed http://www.talafone.com/

RandomGuy48:
Domain for sale


agrabeli:
Thank you @RandomGuy48

PR #849, opened 2021-10-07
Duplicate Test?

RandomGuy48:
https://ooni.org/nettest/telegram
image


RandomGuy48:
Also, I updated the URL into https://www.blackberry.com/ because of #40 & #58

agrabeli:
Good catch! Thank you @RandomGuy48

RandomGuy48:
@agrabeli https://ooni.torproject.org/ is a duplicate too since it redirects to https://ooni.org/ but the former seems to have an issue
image

PR #851, opened 2021-10-14
Removed https://groups.yahoo.com/ & https://kids.yahoo.com/

RandomGuy48:
- https://www.zdnet.com/article/yahoo-groups-to-shut-down-for-good-on-december-15-2020/

- https://news.thewindowsclub.com/yahoo-retire-several-products-61291/


agrabeli:
Thank you @RandomGuy48

PR #855, opened 2021-10-14
Removed http://www.law-lib.utoronto.ca/Diana/

RandomGuy48:
Dead link


agrabeli:
Thank you @RandomGuy48

PR #857, opened 2021-10-28
Moved 3 URLs from global.csv to us.csv

RandomGuy48:
Transferred due to relevance, Rise Broadband and Viasat are commercial Internet services available on US. DC Pages is Washington DC Photo Gallery


agrabeli:
Thank you @RandomGuy48 . The Washington Post and USA Today are meant to be included in the Global test list, so that all OONI Probe users (regardless of country, and including the US) can test them.

RandomGuy48:
I see @agrabeli but should [Rise Broadband](http://risebroadband.com/) be included in the Global test list too? If yes, it should be removed from the [global.csv](https://github.com/citizenlab/test-lists/pull/857/commits/c27ea76504e6e56c4deb905385fcff48f3c46013). If no, it should be removed from the [us.csv](https://github.com/citizenlab/test-lists/pull/857/commits/c26e86f3ffc51ac88e43c2b640fba310ff86e866) to avoid duplicates.

PR #867, opened 2021-11-30
Add https://www.dlsite.com/ to TH list

### LINKED ISSUE #866: www.dlsite.com (and several other Japanese websites) cannot be access from Thailand
As reported on CatDumb on 29 Nov 2021, several Japanese websites cannot be access from Thailand,
saying that it needs "ICP Beian" or the Chinese "Internet Content Provider" license.
https://www.facebook.com/CatDumbNews/posts/4796921403683944 

One of them is https://www.dlsite.com/ (an online store for digital content, games, anime, including non-adults and adults content).

Visitors from some Thai ISPs may get a 451 error message, saying:

"451 Unavailable for Legal Reasons: This request may not be serviced in mainland China because the domain name does not have the required ICP Beian."

The issue can be solved when changing DNS to public DNSs like Google's and 1.1.1.1.

bact:
A Japanese website that reported to require a Chinese Internet Content Provider license in other to view from Thailand. 

This is to close #866 (see details there)


agrabeli:
Thank you @bact !

PR #870, opened 2021-12-08
Updated Canada local list with file-sharing sites ordered blocked via…

sneft:
… court-order


hellais:
Do you have more context around the reason why the block was ordered? I tried accessing some of the link in there and they all seem to contain some strange looking login pages that don't offer much content.

sneft:
Yes, the court has ordered the blocking of a set of IPTV streaming services for violating copyright [1]. Since the court recognizes that these IPTV providers may just switch domains/IPs in response to being blocked, the court has allowed for the list of sites to be updated over time, and these changes I've added in this PR reflect the most recent update from September [2].

It's possible that some of these sites/domains are not functional as the sites move services elsewhere, but by law all major ISPs are required to block the websites I've added to this list. Whether ISPs are actually following these orders is an outstanding research question, so I'd like to collect some data here to try to answer that. 

[1] News coverage here: https://www.cbc.ca/news/entertainment/federal-court-bans-piracy-website-1.5365032
[2] Teksavvy, a Canadian ISP, discusses the law on their blockpage, and links to a PDF of the most recent domain/subdomain/IP list here: http://goldtvblock.teksavvy.com/

PR #875, opened 2022-01-02
Update in.csv (added piratebay.live to the list)

charlie39:
piratebay.live is proxy site for piratebay.org that has been blocked in india for many years, just noticed it is not in the list so updating it.


agrabeli:
Thank you @charlie39 !

PR #876, opened 2022-01-04
Added lapsiporno.info to FI list

pyllyukko:
This page contains criticism about the net filtering in Finland and was
previously blocked by several operators in Finland.

References:

* https://en.wikipedia.org/wiki/Lapsiporno.info
* https://wikileaks.com/wiki/797_domains_on_Finnish_Internet_censorship_list,_including_censorship_critic,_2008


agrabeli:
I'm not sure about merging this PR, as the test lists have a policy for not integrating anything related to CP for ethical and legal reasons. @hellais @sneft can you please review this PR?

pyllyukko:
> I'm not sure about merging this PR, as the test lists have a policy for not integrating anything related to CP for ethical and legal reasons.

Just to clarify that this site has nothing to do with CP except the domain name.

sneft:
This is a very interesting case. I appreciate that this is intended to be a research and advocacy site that highlights important issues like overblocking and miscategorization of web content. However, based on the Wiki, the site does contain links to actual CSAM-hosting websites. For that reason, I don't think we want to include this in our test lists. 

I appreciate that accessing CSAM content vs. accessing content containing links to CSAM content are distinct, but I am concerned we would be increasing the prevalence of actual CSAM-containing URLs in measurement datasets and increasing the possibility of a researcher inadvertently accessing such content.

PR #878, opened 2022-01-24
include OSSRH Maven Central in global list

eighthave:
OSSRH Maven Central is the default library repository for Java and Android projects.  It is a massive collection of open source Java libraries, including a number of circumvention tools.  It requires HTTPS so it could be a candidate for being fully blocked in order block access to specific packages that are hosted there.  It could be consdered something like GitHub or Bitbucket.


agrabeli:
Thank you @eighthave !

PR #879, opened 2022-01-24
Update global.csv

epitome-awardee:
Added Kiwifarms as they are now blocked by ROSKOMNADZOR


agrabeli:
Thank you @epitome-awardee !

PR #889, opened 2022-02-07
Updates to Global list

agrabeli:
I did a bit of clean-up to the Global test list, updating some URLs to HTTPS, and removing a few URLs that are known to (almost always) cause false positives.

A more extensive clean-up and update to the Global test list is needed.


agrabeli:
I opened this PR https://github.com/citizenlab/test-lists/pull/890 to remove duplicates (country vs global lists) and unblock the merging of this PR.

PR #898, opened 2022-02-16
Removed https://www.webspawner.com/

RandomGuy48:
https://www.webspawner.com/ was a dead website transfered to ir.csv due to its blocked in Iran


agrabeli:

PR #899, opened 2022-02-16
Added https://www.webs.com/

RandomGuy48:
Transferred https://www.webs.com/ to global.csv due to global relevance
Also, removed http://www.freewebs.com/ to ph.csv since it redirects to webs.com


agrabeli:
Thank you @RandomGuy48 !

PR #900, opened 2022-02-16
Removed https://extratorrent.cc/

RandomGuy48:
https://extratorrent.cc/ was a dead website transfered to ir.csv & my.csv due to its blocked in Iran and Malaysia (https://techlog360.com/rip-extratorrent/)


agrabeli:

PR #902, opened 2022-02-16
Removed http://www.mate1.com/

RandomGuy48:
http://www.mate1.com/ was a dead website and transferred to de.csv, id.csv & ir.csv due to its blocked in these countries


agrabeli:
Thanks @RandomGuy48 !

PR #904, opened 2022-02-16
Replaced picasa.google.com into photos.google.com

RandomGuy48:
image


agrabeli:
Thank you @RandomGuy48 !

PR #905, opened 2022-02-16
Removed http://www.lingerieatlarge.com/

RandomGuy48:
http://www.lingerieatlarge.com/ was a dead website and transfered to de.csv, ir.csv & sa.csv due to its blocked in these countries


agrabeli:
Thanks @RandomGuy48 ! Since this website is dead, I'd recommend removing it from all test lists (particularly since other websites of this category are already tested globally). While I see some value in testing it in countries where it's blocked (as that can, for example, help suggest how frequently censors update their blocklists), I think that value may be rather limited since the website no longer exists anyway.

RandomGuy48:
I see, I'll compile all these dead websites and create a new PR

PR #907, opened 2022-02-17
Add no112.org URLs

bact:
Add https://www.no112.org/ a signature collection website to propose the end of the lese majeste law (Article 112).

Reported to be blocked since 15 Feb 2022
https://www.matichon.co.th/news-monitor/news_3185821


agrabeli:
Thank you @bact !

PR #910, opened 2022-02-21
Added www prefix to URLs and corrected some categories

Raja063:
#58
> I think that the general rule should be to use the www prefix only when a request to http://example.com does a 30x redirect to http://www.example.com.


agrabeli:
Wow, great work! Thank you @Raja063 !

PR #901, opened 2022-02-16
Removed https://torrentz.eu/

RandomGuy48:
https://torrentz.eu/ was a dead website transfered to ir.csv & my.csv due to its blocked in Iran and Malaysia (https://www.news18.com/news/buzz/rip-torrentz-eu-website-bids-the-world-a-sentimental-farewell-1277997.html)

PR #895, opened 2022-02-16
Removed community.livejournal.com/anorexicqueen/

RandomGuy48:
http://community.livejournal.com/anorexicqueen/ was a purged account transfered to de.csv & ir.csv due to its blocked in Germany and Iran

PR #912, opened 2022-02-23
Removed dead websites

RandomGuy48:
As mentioned in #895 #896 #897 #898 #900 #901 & #905 and removed some more


agrabeli:
THANK YOU @RandomGuy48 ! <3

PR #917, opened 2022-02-24
Patch 2

RandomGuy48:
https://www.facebook.com/projectdhee/ - Facebook page not found
http://www.atimes.com/atimes/South_Asia/DI21Df06.html - dead link
Were already in the global test list:
https://bit.ly/
https://archive.org/
https://mediafire.com/ (lacks www prefix)
https://change.org/ (lacks www prefix)


agrabeli:
Thank you @RandomGuy48

PR #919, opened 2022-02-25
Updated YouTube video URLs & Facebook/Instagram page links

RandomGuy48:
The following YouTube links are checked/visited without loging in to Google.

Update:
I checked them using an alternate Google account today. None of them are private videos. These videos are unavailable anymore.


agrabeli:
Thanks!

PR #924, opened 2022-02-26
Removed http://hrw.org/ on country-specific list

RandomGuy48:
https://www.hrw.org/ is already listed on the [global test list](https://github.com/citizenlab/test-lists/blob/master/lists/global.csv)


agrabeli:
Thanks!

PR #926, opened 2022-02-27
Removed duplicate links in ph.csv

RandomGuy48:
Duplicates from ph.csv
http://raissarobles.com/
http://www.arkibongbayan.org/
http://www.filipinaheart.com/ which redirects to https://www.filipinocupid.com/ that is already present in the list
http://www.gmanews.tv/ which redirects to https://www.gmanetwork.com/news/ that is already present in the list
http://www.pinglacson.com.ph/ which redirects to https://pinglacson.net/ that is already present in the list
http://www.rappler.com/
http://www.sunstar.com.ph/

Duplicates from Global Test List
https://www.arabnews.com/
http://www.cbsnews.com/
https://www.greenpeace.org/
https://www.msn.com/
https://www.telegraph.co.uk/


agrabeli:
Amazing work! Thanks a million!

PR #928, opened 2022-02-27
Update global.csv

rayasharbain:
Found blocked on some ISPs in Jordan, could be more relevant to global list.


agrabeli:
Thanks!

PR #931, opened 2022-03-06
Add rtnews to the global list

hellais:
It seems to be geoblocked in some places and it would be interesting to measure the effect of it.


hellais:
Self merging as it's related to an ongoing incident

PR #933, opened 2022-03-09
Updated URLs in ph.csv

RandomGuy48:
#727 removed _http://codengo.org/_ & _http://philippinecollegian.net/_ but I found their working links: _https://code-ngo.org/_ & _https://phkule.org/_ respectively.

_http://groups.google.com/group/soc.culture.filipino/topics?lnk=sg_ & _http://www.radikalchick.com/_ redirects to _https://groups.google.com/g/soc.culture.filipino_ & _http://katrinasantiago.com/_ respectively.

I also change _http://anakpawis.blogspot.com/_ & _http://magdaloparasapagbabago.blogspot.com/_ into _https://anakpawis.net/_ & _http://magdaloparty.com/_ because the former is unmaintained and I believe the latter where their new domains.

The _http://blogwatch.ph/_, _http://www.liberalparty.ph/_, _http://www.bangsamoro.com/_ & _http://www.undp.org.ph/_ are dead websites which I looked in the internet to find their working domains.

I would like to ask why the URLs provided by Citizen Lab are http and not https? While some of them do not have https protocol, most of them have. I did not include it in this update but I would like to change those with https protocol if it is okay.


agrabeli:
Thank you @RandomGuy48 !

PR #934, opened 2022-03-09
Added new BBC domain to RU test list

agrabeli:
This PR adds a new BBC domain that has been set-up following the blocking of BBC in Russia.


agrabeli:
I'll proceed with merging this so that it can get tested ASAP.

PR #943, opened 2022-03-19
Update global.csv ( Adds raidforums.com)

takipsizad:


hellais:
It looks like this site has been taken down. As a result of that I am not sure it's relevant to have it be tested, given that it's unlikely to be blocked through network filtering.

takipsizad:
it has a chance of it though and yeah site doesnt exist anymore

agrabeli:
Thank you @takipsizad for your contribution! I agree though with @hellais and think that since the domain has been seized, it's probably not worth adding it to the global test list (and having all OONI Probe users worldwide test it) because even if it gets blocked, the site doesn't exist anyway. 

I'll proceed with closing this PR. We welcome any other contributions you may have! <3

PR #944, opened 2022-03-23
Remove https://mitm.watch/

catileptic:
the URL no longer resolves, the A record has been removed


agrabeli:
Thank you @catileptic !

PR #946, opened 2022-03-30
Added global test for starlink.com

nunogt:
Added global test for starlink.com, already banned in some countries due to its role in censorship circumvention and emergency internet access. Currently not tested.


agrabeli:
Thank you @nunogt

PR #948, opened 2022-03-31
Added Zoom and Microsoft URLs to global test list

agrabeli:
This PR includes URLs provided by a community member for the more extensive testing of Zoom and Microsoft (which were reportedly blocked yesterday in Tunisia).


lists/global.csv
```diff
@@ -1505,4 +1505,13 @@ https://www.facebook.com/RTNews,GRP,Social Networking,2022-03-06,OONI,
...
 https://www.starlink.com/,ANON,Anonymization and circumvention tools,2022-03-30,community member,Censorship circumvention and emergency internet access
 https://dailystormer.name/,POLR,Political Criticism,2022-03-18,community member,American far-right website; supposedly blocked in Europe
-https://www.microsoft.com/,COMT,Communication Tools,2022-03-30,OONI,Reportedly blocked in Tunisia but worth testing globally
\ No newline at end of file
+https://www.microsoft.com/,COMT,Communication Tools,2022-03-30,OONI,Reportedly blocked in Tunisia but worth testing globally
+https://teams.microsoft.com/,COMT,Communication Tools,2022-03-31,OONI,
+https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration,COMT,Communication Tools,2022-03-31,OONI,
+https://statics.teams.cdn.office.net/hashed/favicon/prod/favicon-9e2b8f1.ico,COMT,Communication Tools,2022-03-31,OONI,
+https://login.live.com/Me.htm?v=3,COMT,Communication Tools,2022-03-31,OONI,
+https://aadcdn.msauth.net/shared/1.0/content/images/favicon_a_eupayfgghqiai7k9sol6lg2.ico,COMT,Communication Tools,2022-03-31,OONI,
```
hellais, line 1513: Based on the PCAPs I have seen of the microsoft teams app, the only domain used by the app is https://teams.microsoft.com/.

I suggest dropping all other URLs and keeping just that.

lists/global.csv
```diff
@@ -1505,4 +1505,13 @@ https://www.facebook.com/RTNews,GRP,Social Networking,2022-03-06,OONI,
...
+https://explore.zoom.us/docs/aboutAssets/img/ZoomLogo.png,COMT,Communication Tools,2022-03-31,OONI,
+https://st1.zoom.us/zoom.ico,COMT,Communication Tools,2022-03-31,OONI,
+https://st2.zoom.us/zoom.ico,COMT,Communication Tools,2022-03-31,OONI,
+https://st3.zoom.us/zoom.ico,COMT,Communication Tools,2022-03-31,OONI,
```
hellais, line 1517: Are all of these used by the zoom app for communication?

lists/global.csv
xhdix: Required for use on the web. (login)

(PCAPs were probably created after login or in use. In this case, some important and basic addresses may not be visible.)

lists/global.csv
xhdix: These are CDNs for using the service on the web.

Maybe keeping only one item on the global list is enough? I did the same for other services. e.g.: https://github.com/citizenlab/test-lists/blob/master/lists/global.csv#L1366-L1369

PR #963, opened 2022-04-22
update aihrc.org.af HTTP to HTTPS in AF list

xhdix:


xhdix:
Oh! thanks! It's interesting that I did not see it. :man_facepalming: 
done!

PR #971, opened 2022-05-12
SearX, PrivacyTools Search are MIA

arky:
Removing SearX and PrivacyTools Search service which are discontinued and offline.


agrabeli:
Thank you @arky !

PR #985, opened 2022-05-19
Remove all gov.br websites

gusgustavo:
It's very unlikely that the Brazilian government would ask to the Brazilian ISPs to block these websites. 
I'm removing all .gov.br websites of the BR test list.


agrabeli:
Indeed! Thank you @gusgustavo ! <3

ramakrishnansr:
I agree that having > 1,500 .gov.br websites is a stretch, but I still think it's valuable to have some .gov.br (perhaps the most popular or most important) in the BR list. This is because I think people outside Brazil might test the BR test list as well, and in that case it's useful to record whether Bralizian government websites are reachable.

gusgustavo:
I kept three BR gov websites in the test list:

```
http://www.tse.jus.br/
http://www.blog.saude.gov.br/
http://www.tre-rs.jus.br/
```

I can add the main BR gov: `https://www.gov.br/` in a new PR.

PR #998, opened 2022-05-31
Updated links and removed duplicates

RandomGuy48:


agrabeli:
@RandomGuy48 thank you! <3

PR #1000, opened 2022-06-07
Updated AZ, RU, and Global lists

agrabeli:
In this PR I've added AZ news media websites (to the AZ and RU lists) which are reportedly blocked in RU. I've also added RIA Novosti to the Global list, as it's reportedly blocked in Azerbaijan and blocked in EU countries.


ramakrishnansr:
@agrabeli It looks like `http://ria.ru/` is in the tj.csv and md.csv local lists? What is the protocol for duplicating URLs between local and global lists?

agrabeli:
@ramakrishnansr wow! Thanks for the quick response. :) The process is that if a URL should be prioritized for global testing, we remove it from local lists and just keep it in the global list. Given that ria.ru is reportedly blocked in AZ and for sure blocked in LT (and likely many other EU countries), I propose we keep it in the global test list for global testing. What do you think? If you agree, I can proceed with removing it from the other lists.

ramakrishnansr:
That sounds good to me! Thanks @agrabeli!

agrabeli:
As this PR has been reviewed by @ramakrishnansr , I'll proceed with merging so that we can start getting measurements on the blocking of AZ news media sites in RU asap.

PR #1002, opened 2022-06-09
Possible miscategorisation of www.answering-islam.org in MY list?

sitinurliza95:
Found that https://www.answering-islam.org/ is more likely a religion-related website, rather than a news media site. Could this be a possible miscategorisation?


agrabeli:
Indeed, thank you @sitinurliza95 !

PR #1003, opened 2022-06-09
Update dz.csv

BoFFire:
Adding :

- change.org
- patreon.com
- gofundme.com

All blocked by all algerian ISPs.


agrabeli:
Thank you @BoFFire ! `change.org` and `patreon.com` are already included in the Global test list, which means that they are amongst the URLs that are tested worldwide (including Algeria). Similarly, `gofundme.com` should probably be added to the Global test list and measured globally too. I'll proceed now with adding it to the Global test list (and then I will close this PR).

agrabeli:
@BoFFire I just opened a PR adding gofundme.com to the Global test list (so that it can get tested globally, including Algeria). The other 2 URLs are already included in the Global test list. I'll therefore proceed with closing this PR. Thanks for notifying us of these URLs! <3

PR #1016, opened 2022-06-24
fix: pmdb.org.br -> www.mdb.org.br

bassosimone:
A user mentioned to us that the political party changed name and
website. This diff updates to the new website.


agrabeli:
Thank you!

agrabeli:
(I changed the category from MISC to POLR because it's a political party website and should be categorized as POLR)

Issue #979, opened 2022-05-15
www.phenoelit.org is gone


davidak:
The domain www.phenoelit.org is part of your list, but it is not reachable. I e-mailed FX, president of phenoelit, about the issue 15 days ago, but got no response.

https://github.com/citizenlab/test-lists/blob/84d6a01c1fb797d437f52f8ac4a62fc4176c4fbd/lists/global.csv#L795

https://downforeveryoneorjustme.com/phenoelit.org?proto=http&www=1
https://down.com/

So maybe remove it to prevent false-positives.

(You can still access the website at http://194.150.168.69/)

Issue #824, opened 2021-09-21
Telegram Duplicate URLs on Global Test List


Harpy074:
Is there a reason for having multiple links for telegram.org?
![image](https://user-images.githubusercontent.com/88121480/134144727-e903d79d-7b40-4373-b164-b510281783db.png)
Is it okay to remove these duplicates?

Harpy074:
If YES, maybe someone should make a PR to remove them, I guess.

If NO, does it mean that one should not also update "old domain that redirects you to its new domain" like https://postimage.io/ that redirects to https://postimages.org/?
- If so, then would this be considered as duplicates:
![image](https://user-images.githubusercontent.com/88121480/135445375-243cb50f-0920-45ac-af89-5ce91c3196b7.png)
_(Photo above came from global.csv and photo below came from ir.csv)_
![image](https://user-images.githubusercontent.com/88121480/135445285-26cbd9aa-1e79-4745-93f1-982bddc83ef0.png)
- Or should both links be in the global list (global relevance) or in Iran test list (country-specific)?

_P.S.: I'm not very confident with my grammar but I hope you get my point_
_P.P.S.: telegra.ph (1380) was not included, I did not remove it from the photo because the Telegram links are right next to each other (1376-1381)_

Vort:
> Is there a reason for having multiple links for telegram.org?

I think, yes.
Some dumb government/ISP may block one of the addresses, but forget other ones and test will show it.

bact:
There were a lot of cases that variants of a blocked URL are not similarly blocked.
For example, http://www.example.com may got blocked, but not http://example.com or https://www.example.com.

PR #1018, opened 2022-07-03
Updated test lists

Raja063:


Raja063:
#824: Question about _postimage_
> should both links be in the global list (global relevance) or in Iran test list (country-specific)?

Raja063:
#824: Question about _postimage_
> should both links be in the global list (global relevance) or in Iran test list (country-specific)?

Raja063:
#979: Question about _Phenoelit_

> The domain [www.phenoelit.org](http://www.phenoelit.org/) is part of your list, but it is not reachable...
So maybe remove it to prevent false-positives.
(You can still access the website at http://194.150.168.69/)

PR #1028, opened 2022-07-23
fix: category of medium.com

developStorm:
medium.com is more of a general blogging platform.


sloncocs:
Thank you @developStorm for the contribution!  <3

PR #1031, opened 2022-07-25
Update mx.csv

carlosm2:
remove 'http://base.espora.org/tiki-index.php' doesn't exist anymore
remove 'http://espora.org/comitecerezo/' and changed for the actual site
remove 'http://espora.org/revueltas/' doesn't exist anymore
remove 'http://imc-yucatan.espora.org/' doesn't exist anymore
add https://espora.org/pages/About.html autonomous server


sloncocs:
Thank you @carlosm2 for your contribution, you also changed the 'espora.org' category from HACK, Hacking Tools to POLR, Political Criticism. Could you please elaborate a bit on that? According to the 'about' section on the website I would still suggest the HACK category to be more relevant.

carlosm2:
Hi @sloncocs, first thanks for reviewing my PR 👍🏾, to clarify the change of category I looked for similar projects, that are autonomous servers (https://riseup.net/en/security/resources/radical-servers) and I realized that autistici, and noblogs are in the "HOST" category, then perhaps it is convenient to change the category, instead of POLR to HOST. By the way, I was part of espora.org for a few years and no HACK things were shared there, actually most of the things in espora are hosting social movements. Thanks

PR #1032, opened 2022-07-25
Deleted duplicates, added media 

sloncocs:
Deleted duplicates, and added 57 local independent media, most of them were blocked once or more since February 24th


agrabeli:
Thank you @sloncocs ! <3

PR #1038, opened 2022-07-28
Updated test-list, Georgia

sloncocs:


agrabeli:
Thank you @sloncocs for your contributions! <3

A bit of feedback:

- I have updated some URLs to HTTPS (https://github.com/citizenlab/test-lists/pull/1038/commits/e648b1fd0fa9fe3e494545f991aefe2f6a192d51) -- the ones that support HTTPS. Generally, we recommend copying and pasting a URL from a browser (when adding it to a test list) so that users measure a URL in the same format as it would appear if accessed from a browser. 
- Are you able to access `http://detals.net/` from a browser? I get an NXDOMAIN error. If not, we can remove this URL from the list (and then proceed with merging). 

Thanks again! <3

sloncocs:
Thank you! I copied most of them from another website, and, apparently haven't checked all of them through the browser! Thank you for updating HTTPS, the detals.net doesn't open from my side either, so I suppose we can remove it from the list. 

I also had a question about putting some organisations within the 'Terrorism and Militants' (MILX) category, e.g., in this list all organisations related to Abkhazia region have MILX code, most likely because Abkhazia is a separatist state since 1992 which is not recognized by anyone except Russia. However, the labelled organisations are not only governmental or political ones but also media. And I was confused if we should label them as media or leave with MILX supposing that they are supported by separatist movements (e.g., in this case by Russia)?

agrabeli:
@sloncocs thanks for your feedback, I just removed `detals.net`. 

Great question about the MILX category and sites related to Abkhazia! While I can't say with certainty, I generally think that they should probably be treated on a contextual, case-by-case basis. For example, if it's a website Abkhazia-related site that specializes on sharing news, it could be categorized as `NEWS`. If it's a related site that shares political views, it could be categorized as `POLR`, etc. 

But I recognize that what is labelled as `MILX` can sometimes be quite political in itself, as what may be viewed as a human rights movement by some, may be viewed/labelled as a terrorist action by others. We generally tend to use the `MILX` category for what is internationally viewed/labelled, for example, as a separatist movement. But I realize that a lot of this can be tricky, and would appreciate your thoughts, as well as the feedback from others in the region.

PR #1052, opened 2022-08-10
Fix/dupe categories

hellais:
In this PR I address the issue of the same identical URL having different category codes inside of different country list.


hellais:
I am going to proceed with landing this as it's a requirement to https://github.com/ooni/api/pull/300

PR #1059, opened 2022-08-12
Updated LB and Global lists

agrabeli:


sloncocs:
Thank you @agrabeli! <3

PR #1073, opened 2022-08-16
add z-library to India list

d1vyank:
https://entrackr.com/2022/08/exclusive-z-library-sites-blocked-in-india-following-court-order/


Te-k:
Considering that the Z Library start to be blocked in many places, it could be interesting to move Zlib urls in the global list

d1vyank:
Thanks for pointing that out, @Te-k. I've done so here: https://github.com/citizenlab/test-lists/pull/1162

PR #1111, opened 2022-09-14
Added TikTok endpoints to AZ test list

agrabeli:
We're receiving reports that the TikTok app may have been blocked in Azerbaijan today. This PR therefore adds TikTok endpoints to the AZ test list. 

I've refrained from adding these endpoints to the Global test list, as it's quite a long list of endpoints (whereas the AZ list -- where the blocking is being reported -- is short). We could potentially remove these endpoints from the AZ list (and potentially add some or all of them to the Global list) once we've collected enough data over the next weeks to investigate this case.


agrabeli:
I'll merge this PR so that these URLs can get prioritized for testing (given the urgency).

PR #1128, opened 2022-09-28
Updated Global test list

agrabeli:
This PR adds URLs (shared by community members) that are reportedly blocked in Iran, but worth testing globally.


bassosimone (approved):
🐳

PR #1019, opened 2022-07-05
Add online articles on MACC chief commissioner

sitinurliza95:
There were signs of blocking on a blog with articles concerning alleged share trading controversy involving Malaysian Anti-Corruption Commission chief commissioner Azam Baki here: http://edisisiasatmy.blogspot.com/.

Hence in relation to this, adding these 2 news articles regarding the case:
https://www.theins.news/postview/757-business-ties-among-macc-leadership-how-deep-does-it-go-part-1				
https://www.theins.news/postview/758-business-ties-among-macc-leadership-how-deep-does-it-go-part-2

PR #1130, opened 2022-09-29
Add a few links

sitinurliza95:
Following up on https://github.com/citizenlab/test-lists/pull/1019, resolved conflicts with latest urls


sitinurliza95:
Propose to update the MY list as follows:

- Change: Muda's website: https://partimuda.my/ -> https://muda.my/

- Remove:
1. https://www.gayasiantheater.com/dispatcher/frontDoor?theaterId=43045&genreId=102&locale=en/ (Down since 6/2022)
2. https://bangkokbangers.com/ (Down since  12/2021)
3. http://www.suarakeadilan.com/ (Down since 2/2022 or earlier); also I think it's replaced by https://www.suarakeadilan.my/, which is already in the list
4. http://www.terrorismfiles.org/organisations/harakat_ul-mujahideen.html (Domain sold)

There's also websites that are only recently down, but I'll keep them for now
1. http://virtualfriends.net/ (Down since 8/2022)
2. http://ww82.chedet.cc/ (Down since 9/2022)

agrabeli:
Thank you @sitinurliza95 ! <3

PR #1134, opened 2022-09-30
Removed Expired domain

arky:
Removing norodomranariddh.org domain from KH test list.


sloncocs:
Thank you for your contribution  <3

PR #1140, opened 2022-09-30
Remove www.hivandhepatitis.com from Global

arky:
Remove inaccessible and offline website www.hivandhepatitis.com from Global

https://explorer.ooni.org/chart/mat?test_name=web_connectivity&input=http%3A%2F%2Fwww.hivandhepatitis.com%2F&since=2022-09-01&until=2022-10-01&axis_x=measurement_start_day


sloncocs:
Thank you for your contribution <3

PR #1176, opened 2022-11-01
Adds more blocked websites in France

Te-k:
Hi, 

Here is an additional list of blocked domains in France (I removed `wearechange.org` because it is already in global)


sloncocs:
@Te-k thank you so much for this amazing contribution! I have minor suggestions for some of the URLs: 

1. https://allomegavideo.com/; https://sekdrive.net/ show DNS_PROBE_FINISHED_NXDOMAIN error, which means that their domains are not active at this moment for some reason, as we don't know if they will be available anytime soon, I am not sure it's feasible to add them to the list
2. alloseven.com domain is put up for sale (https://www.dynadot.com/forsale/alloseven.com?drefid=2071) so I would suggest removing it from the commit
3. https://rt.com/ is on www domain, and https://www.rt.com/ is already added to the global list and therefore is being tested in France too, so I would suggest removing it from your commit

Please let me know if you have any doubts or questions about these, otherwise happy to merge the PR as soon as they are addressed.

Te-k:
Hi @sloncocs , thanks for the feedback, I have made the changes you mentioned.

sloncocs:
Thank you @Te-k! Merging it now

PR #1187, opened 2022-11-25
Fix miscelaneous typo in the list of categories

abitrolly:

PR #1189, opened 2022-11-25
Non-political sites blocked from outside

abitrolly:
Precedence that could be used to attack net neutrality should be tracked as well.


abitrolly:
Ok. This needs #1187 to be merged first.

agrabeli:
Thanks @abitrolly !

Tobie1992 (changes requested):
Good

PR #1195, opened 2022-12-07
Adding dawnmena.org to Saudi Arabia test list

rayasharbain:
Added DAWN's website which is possible blocked in KSA


sloncocs:
Thank you for the contribution @rayasharbain! <3

PR #1199, opened 2022-12-13
lists/ir.csv: fix handle link

alanorth:
This was supposed to be testing resolution of the hdl.handle.net do- main itself, not what the Handle link resolves to (Handles are like DOIs, where a persistent identifier points to a resource on another domain). We had a case of hdl.handle.net being blocked in Iran (not sure if still the case).

This test was erroneously updated to point to the resolved resource (I suspect someone used a script to update the list, because the Handle service would have returned an HTTP redirect). Furthermore, a trailing slash on this URL is not even valid (returns HTTP 404).

See: https://github.com/citizenlab/test-lists/pull/323
See: 34cac1b5b881404faf21ee9bf11f33945bd77129


sloncocs:
Thank you @alanorth!

PR #1201, opened 2022-12-16
Added TikTok endpoints to JO test list

sloncocs:
We're receiving reports that the TikTok app may have been blocked in Jordan today. This PR, therefore, adds TikTok endpoints to the JO test list.


hellais:
LGTM

PR #1213, opened 2023-02-01
chore(it.csv): zap parked/nonexisting domains

bassosimone:
This diff deletes a bunch of domains that we found to be parked or nonexisting a while ago when testing web connectivity.

All these domains are part of the Italian test list.

They all have some residual censorship, but we've been favouring existing websites recently, so I think this diff makes sense.

Closes https://github.com/ooni/probe/issues/2316

Closes https://github.com/ooni/probe/issues/2315

Closes https://github.com/ooni/probe/issues/2314

Closes https://github.com/ooni/probe/issues/2312

Closes https://github.com/ooni/probe/issues/2302

Closes https://github.com/ooni/probe/issues/2308

Closes https://github.com/ooni/probe/issues/2304

Closes https://github.com/ooni/probe/issues/2310


sloncocs:
Thank you @bassosimone! 

I found that one of the services moved to a new domain: http://ww1.cineblog01.fm/ --> https://www.cb01.in/

In the case of manual review, the new domain would be added instead of removed by the researchers. While it may be not so critical for file-sharing services, it is very important for such categories as NEWS, POLR, HUMR, and ENV, so I would suggest keeping a list of removed parked domains for some categories and reviewing it manually from time to time to avoid losing important resources.

I will merge this PR now and add https://www.cb01.in/ to the Italian test list.

PR #1221, opened 2023-02-11
Specify utf-8 encoding when loading csv's in lint-lists.py

DaemonUmbra:
Allows the script to process non-latin characters on Windows, which defaults to cp1252


hellais:
Thanks!

PR #1222, opened 2023-02-13
Update hk.csv

constantinedev:
Add the status column to the first column,
scan with all normal network request with using:
python requests

Mark this down kind of the URI are getting Timeout unlimited from the responed server.
All make the mark to [Err]
All the response with 200 mark at [ARV]


sloncocs:
@constantinedev could you please let us know which server did you use to test the accessibility of the URLs? and if there was any script, could you please share a link where we could check it out?

PR #1225, opened 2023-02-15
Update lists/pk.csv to add Wikimedia Commons

azadi:
Add commons.wikimedia.org to lists/pk.csv.


azadi:
CI is failing with:

```
lists/pk.csv (line 625): Duplicate URL between Local List and Global List
```

Which is correct; however, the reason for adding it to the pk.csv list -- as per what I remember and sorry for not being up-to-date -- is that in case someone selects the country-specific list and not the global one, that they can still run the test.

Is that assumption still correct?

sloncocs:
hi @azadi! thank you for your contribution! 

The global test list is being tested by all OONI Probe users in all countries, so if there is a duplicate between a country-specific list and a global list, this would mean that the URL is being tested twice. 

The only opportunity for the user to test only a country-specific list would be to copy the whole list and create an OONI Run (https://run.ooni.io/) link to test it. As this is not a very common case, I would suggest not adding commons.wikimedia.org to lists/pk.csv, and leave it only on the global list.

azadi:
Thanks for the clarification @sloncocs!

PR #1230, opened 2023-02-21
Updated my.csv

sitinurliza95:
Updated MY test list with latest political parties websites, news media websites and websites found blocked in Malaysia. Also removed inactive websites which were down for a while.


sitinurliza95:
Hi @sloncocs it's a website of an political opposition party in Malaysia. According to Wayback Machine here https://web.archive.org/web/20230302130732/https://tanahairku.my/ it was still active until 2 March recently, so I think it's good to keep it on the test list for a while. I will monitor it and see if it's need to be removed later.

sloncocs:
Thank you, we'll leave it as it is then. Merging the PR now!

PR #1236, opened 2023-02-22
[th] Add two soccer live streaming websites

bact:
They probably got banned recently


agrabeli:
Thank you @bact, LGTM too! Merging.

PR #1250, opened 2023-03-19
Add more URLs to the test-lists from mining OONI measurements

hellais:
These are sites that OONI users chose to test, which aren't in the lists


anadahz:
LGTM!

PR #1260, opened 2023-03-31
[th] Add www.vote62.com - election monitoring

bact:
- www.vote62.com - crowdsource election monitoring platform


sloncocs:
thank you @bact!

PR #1275, opened 2023-04-24
Add news websites mentioned in EU sanctions

anadahz:
https://www.consilium.europa.eu/en/infographics/eu-sanctions-russia-ukraine-invasion/


anadahz:
Given that these websites have been mentioned in EU sanctions for blocking, will it make more sense to move them to the global list?

sloncocs:
@anadahz thanks, that's a great suggestion, I deleted the domains from the country-specific test lists and will merge your PR now.

anadahz:
@sloncocs Great, thank you!

PR #1279, opened 2023-05-01
Removed defunt CNRP7 and RFA entry

arky:
Removing defunct CNRP7 domain name. 

Removing RFA.org which is already in Global list.


arky:
Clean-ups related to https://github.com/ooni/probe/issues/2452

sloncocs:
Thanks @arky!

PR #1282, opened 2023-05-04
[th] Add Voting Station Locator websites

bact:
For election on 14 May and advanced voting on 7 May 2023:

General Election
https://boraservices.bora.dopa.go.th/election/enqelection/ 

Local election
https://stat.bora.dopa.go.th/Election/enqelectloc/


sloncocs:
Thank you @bact! :)

PR #1286, opened 2023-05-05
Drop tor directory authorities from global list

hellais:
These are already being tested as part of the tor reachability test. No need to duplicate coverage.


sloncocs:
Thank you @hellais!

PR #1289, opened 2023-05-05
[th] Add We Watch and ECT report sites (Thai Election)

bact:
- https://wewatchthailand.org/
- https://www.electionwatchth.org/
- https://ectreport.com


sloncocs:
Thanks Art!

PR #1295, opened 2023-05-17
Update tl.csv

sitinurliza95:
Adding a few links based on new iMAP partner for Timor Leste. Would appreciate it if you can merge by 18 May as we will be having an online local workshop. We will also update this after the workshop if there are additional input from the participants.


sloncocs:
@sitinurliza95, thank you for the contribution!

I have minor suggestions:

1. https://aes.exchange/ seems to be a dead domain, it resolves in 404 error on my network, is it different on yours? If not, I would suggest removing it from the list
2. http://bankami.blogspot.com/; http://wfauzdin.blogspot.com/; http://alditta.blogspot.com/; http://kedahlaniie.blogspot.com/; http://penburukonline.blogspot.com/; http://islamsyiah.wordpress.com/; http://abu-yusufmalizi.blogspot.com; http://forum-haksesuk.blogspot.com/; http://matadalaninstitute.blogspot.com/ have an HTTPS version of the website, so I think it's better to add HTTPS URLs
3. some URLs end in '#', could you please delete '#' to avoid duplicates in the list? (if later someone else adds the same domains without #, our script won't see that they are already on the list 
4. same for https://www.thediliweekly.com/tl/; http://jsmp.tl/en/; https://conselhoimprensa.tl/en/; https://www.tic.gov.tl/tp, could you please update them to 'https://www.thediliweekly.com/', 'http://jsmp.tl/', 'https://conselhoimprensa.tl/', 'https://www.tic.gov.tl/' accordingly?
5. https://neonmetin.info/buletin/page/333/ is on HTTPS, so we can limit the URL to 'https://neonmetin.info/'. Also, it resolves DNS_PROBE_FINISHED_NXDOMAIN, so I am not sure we want to add it to the list
6. https://fongtil.org.tl/index.php/pt/ is on HTTPS, so we can limit the URL to 'https://fongtil.org.tl/'

Otherwise, ready to merge :)

PR #1300, opened 2023-05-26
Add njalla to global.csv

gusgustavo:
Adding https://njal.la/ to global.csv file.


sloncocs:
Thanks @gusgustavo! <3

PR #1306, opened 2023-06-13
Update global.csv

gusgustavo:
Add more reproductive rights websites.


sloncocs:
thanks @gusgustavo! I deleted duplicates and updated categorisation to XED for most of the websites as they are focused on abortion and sexual rights

PR #1313, opened 2023-06-20
Add https://hub.docker.com/ to CN.csv

Lanius-collaris:
found SNI filtering in China


sloncocs:
@Lanius-collaris thanks for your contribution!

PR #1361, opened 2023-07-26
DE: Add streaming sites

marek22k:
According to my research, these pages are blocked by some German ISPs via DNS hijacking.


sloncocs:
@marek22k thanks for the contribution! I would avoid adding more PORN websites to the global test list, it already includes a lot of such content, and at the same time overloading it with one specific category will limit our capacity to test other websites in different countries. 

Same for https://dn42.cc/, I would leave 1-2 domains for testing in the global test list, and avoid adding .eu and .de domains to the global test list. 

Could you please update this PR considering these two suggestions?

marek22k:
> @marek22k thanks for the contribution! I would avoid adding more PORN websites to the global test list, it already includes a lot of such content, and at the same time overloading it with one specific category will limit our capacity to test other websites in different countries.

I had not considered that. Thanks for the hint!

> 
> Same for https://dn42.cc/, I would leave 1-2 domains for testing in the global test list, and avoid adding .eu and .de domains to the global test list.

I summarized it on dn42.dev, which is probably the most used mirror.

sloncocs:
Thanks so much @marek22k! I am merging this PR now

PR #1401, opened 2023-08-10
Update tails website

marek22k:
As announced in https://tails.net/news/new_domain/index.en.html, Tails now has its own domain.


sloncocs:
Thanks @marek22k!

PR #1410, opened 2023-08-21
Remove PNN the media which had merged into PTSNews

yanyiyi:
Remove PNN the media which had merged into PTSNews


sloncocs:
Thanks @yanyiyi!

PR #1411, opened 2023-08-22
Update iMAP test list

sitinurliza95:


sitinurliza95:
Added https://sokong.org/ to my.csv

sloncocs:
hi @sitinurliza95! Thanks for the contribution! 

Could you please clarify why did you change `https://www.malaysia-today.net` to `https://malaysia-today.net`? For me the former redirects to www.malaysia-today.net

Same way, `https://samudera.my/`and `https://agendadaily.com/` redirect to `https://www.samudera.my/` and `https://www.agendadaily.com/`, so I would suggest changing them to www domains

sitinurliza95:
Hi @sloncocs the reason is because some ISPs in Malaysia only block them without "www" and would leave the links with "www" unblocked

sloncocs:
@sitinurliza95 oh, thanks for this clarification, it's very important! could we please include both of them (www version and domain without www) in the list?

sitinurliza95:
I think that's up to you/Citizenlab but wouldn't it make the list a little too long? In general, ISPs that do block the "www" version would also block the ones without "www", hence why I kept only the versions without.

PR #1419, opened 2023-08-29
Re-add use-application-dns.net - explicit signal for DNS filtering

R6CG:
With the news that [Firefox plans to start enabling ECH](https://groups.google.com/a/mozilla.org/g/dev-platform/c/uv7PNrHUagA/m/BNA4G8fOAAAJ) in the next releases, I wanted to suggest people check censorship measurements for [use-application-dns.net](https://support.mozilla.org/en-US/kb/canary-domain-use-application-dnsnet), because that canary domain is meant to disable DoH, and ECH depends on DoH.

use-application-dns.net had originally been added in #504 2019-09-19, but then it was deleted in #727 (a dead-domain check) on 2021-02-23.

Maybe prune-dead-urls.py was run on a network that reported NXDOMAIN for use-application-dns.net, since that is the signal that is supposed to signal DNS filtering is in place?

### DIFF
--- a/lists/global.csv
+++ b/lists/global.csv
@@ -1653,3 +1653,4 @@ https://pulse.internetsociety.org/,IGO,Intergovernmental Organizations,2023-07-0
 https://proxy.i2phides.me/,ANON,Anonymization and circumvention tools,2023-07-05,test-lists.ooni.org contribution,Popular I2P Inproxy
 https://www.y2mate.com/,MMED,Media sharing,2023-08-08,test-lists.ooni.org contribution,Youtube video download site
 http://cdns.grindr.com/,LGBT,LGBT,2023-08-09,test-lists.ooni.org contribution,one of the www.grindr.com endpoints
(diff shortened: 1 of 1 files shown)


Lanius-collaris:
The canary domain only applies to users who have DoH enabled as the default option. It does not apply for users who have made the choice to turn on DoH by themselves.
https://support.mozilla.org/en-US/kb/canary-domain-use-application-dnsnet

dnscrypt-proxy also blocks `*.use-application-dns.net` .
https://github.com/DNSCrypt/dnscrypt-proxy/issues/1205
https://github.com/DNSCrypt/dnscrypt-proxy/blob/92e842126d40f6fcb919bce72983748ddb34dac9/dnscrypt-proxy/plugin_firefox.go

bassosimone (approved):
I think we should merge to collect data about which networks _seem_ to block use-application-dns.net. For the analysis, I think we should limit ourselves with analyzing queries generated by getaddrinfo, which _should_ be the automatically configured resolver. But we _should_ also keep in mind that, under specific setups, (e.g., `dnscrypt-proxy` and possibly others), we could get false positives (i.e., the block depends on the local user's configuration and not on the network).

🐳

mScott224 (approved):

PR #1422, opened 2023-09-01
Update URL of verbot-aufheben.org, the previous URL is now for sale

toddy15:


sloncocs:
Thanks for your contribution @toddy15!

PR #1469, opened 2023-10-31
Add jabber.ru

marek22k:
jabber.ru was the victim of a MitM attack (see https://notes.valdikss.org.ru/jabber.ru-mitm/). Furthermore, some ISPs tend to block Russian sites (I know this from German universities, for example). I would therefore find it very interesting to see to what extent jabber.ru is also blocked.


sloncocs:
Thanks @marek22k! I am merging the PR

PR #1471, opened 2023-11-05
Add reported blocks for GB

wheresalice:
Adds some of the more recent blocks from https://wiki.451unavailable.org.uk/wiki/Main_Page and https://en.wikipedia.org/wiki/List_of_websites_blocked_in_the_United_Kingdom as well as recent news coverage


sloncocs:
Hi @wheresalice! Thanks so much for your contribution! 

I have a few minor comments: 

* `https://solarmovie.to/` redirects to `https://ww4.solarmovie.to/`, I would suggest adding the `https://solarmovie.to/` to the test list 
* `https://sanctioned-suicide.org/` seems to result in a DNS_PROBE_FINISHED_NXDOMAIN error, I would suggest removing it from the test list 

Could you please update your PR accordingly? As soon as you update it, I would be happy to merge it

PR #1479, opened 2023-11-17
Removed duplicate URLs

RandomGuy48:
The website jw.org is already present in the global list so I removed them from az.csv, kz.csv, ru.csv, sg.csv, ua.csv, uz.csv and ve.csv
I also changed http://kickassclassical.com/ into http://www.kickassclassical.com/, and https://www.worldrtd.net/ and http://www.worldrtd.net/ into https://wfrtds.org/


sloncocs:
Thanks so much @RandomGuy48!

PR #1483, opened 2023-11-20
Update cd.csv

sloncocs:


agrabeli:
Thanks @sloncocs! For the sites that are "under maintenance" (such as political party websites), please write "Site currently under maintenance" in the "Notes" column so that we can remember to remove them (if they are not restored by the elections or in the next few months). 

We could also consider excluding sites that are not currently operation (and under maintenance).

PR #1486, opened 2023-11-24
global.csv: update OCSP URL of Let's Encrypt

Lanius-collaris:


sloncocs:
Thanks @Lanius-collaris! 🙏

PR #1559, opened 2024-01-09
Updated ph.csv and removed Facebook & YouTube links that are no longer available in other country lists

RandomGuy48:


sloncocs:
Thanks @RandomGuy48! 🙏

PR #1577, opened 2024-01-15
Removed http://www.solicitorsfromhell.com/

RandomGuy48:
Dead link. Last known active: https://web.archive.org/web/20221128122611/http://solicitorsfromhell.com/


sloncocs:
Thanks @RandomGuy48 🙏

PR #1582, opened 2024-01-15
Removed http://triviasecurity.net/

RandomGuy48:
Dead link. I looked at the [Wayback Machine](https://web.archive.org/) and there is a large gap of archiving in this site, it is last known working on 2014.


sloncocs:
Thanks @RandomGuy48! 🙏

PR #1584, opened 2024-01-15
Removed http://www.pokerpages.com/

RandomGuy48:
https://web.archive.org/web/20140511073756/http://www.pokerpages.com:80/articles/archives/announce01.htm


sloncocs:
Thanks @RandomGuy48! 🙏

PR #1588, opened 2024-01-16
Removed http://www.thehacktivist.com/

RandomGuy48:
Last known active: http://web.archive.org/web/20090429065456/http://www.thehacktivist.com/


sloncocs:
Thanks @RandomGuy48! 🙏

PR #1591, opened 2024-01-16
Removed http://www.positive.org/

RandomGuy48:
Dead link. No known new or alternative website of Coalition for Positive Sexuality.


sloncocs:
We see `http://www.positive.org/` still being blocked in multiple countries, e.g., Iran, Indonesia, Saudi Arabia, so I would leave it on the list for the time being.

RandomGuy48:
I encountered the same issue in #896 #897 #898 #900 #901 &
#905 and @agrabeli recommended to remove it from all test lists.
> Since this website is dead, I'd recommend removing it from all test lists (particularly since other websites of this category are already tested globally). While I see some value in testing it in countries where it's blocked (as that can, for example, help suggest how frequently censors update their blocklists), I think that value may be rather limited since the website no longer exists anyway.

PR #1598, opened 2024-01-17
Removed http://www.metacafe.com/

RandomGuy48:
Dead link. Its alternative website: http://videoshub.com/, is also not working.
https://en.wikipedia.org/wiki/Metacafe


sloncocs:
Thanks @RandomGuy48 <3

PR #1605, opened 2024-01-19
Removed http://www.on-instant.com/

RandomGuy48:
Lat known active: https://web.archive.org/web/20140112070827/http://www.on-instant.com/
image


RandomGuy48:
It seems that some ASNs in Cambodia and Indonesia have recently stop blocking _http://www.on-instant.com/_. Nevertheless, I would propose to keep it on the list for the time being to observe these changes.

PR #1608, opened 2024-01-19
Removed http://www.ruf-ch.org/

RandomGuy48:
Domain expired. No known new or alternative website.
https://web.archive.org/web/20230607065819/http://www.ruf-ch.org/


sloncocs:
Thanks so much @RandomGuy48! Merging the PR now.

PR #1461, opened 2023-10-20
Update on ph.csv

RandomGuy48:
SUMMARY OF CHANGES

RESTORED
https://npc-party.org/ (Removed at #952, formerly http://npcparty.org/)
https://prolife.org.ph/ (Removed at #1021, but URL is now working)
https://taskforcedetainees.net/ (Removed at #932, formerly http://www.tfdp.net/)
https://lgbtchamberph.org/ (Removed at #932, formerly http://lgbtph.org/)
https://esp.org.ph/ (Removed at #932, but URL is now working)

REMOVED
http://pankwire.com/ (Dead Website)
http://xtube.com/ (No Longer Available)
http://yaptube.com/ (Dead Website)
http://whoreslag.com/ (Dead Website)
http://theyxxx.com/ (Dead Website)
http://girlsongirlstube.com/ (Redirects to Chinese Site)
http://koalaporn.com/ (Redirects to Chinese Site)
http://pornopantry.com/ (Dead Website)
http://tgutube.com/ (Dead Website)
http://teeztube.com/ (Dead Website)
http://nuuporn.com/ (Dead Website)
http://lcgirls.com/ (Dead Website)
http://cumfox.com/ (Dead Website)
http://gratisvideokijken.nl/ (Dead Website)
http://porndisk.com/ (Dead Website)
http://muslimmindanao.ph/ (Dead Website)
http://www.drugpolicy.org/blog/google-capitalizing-dutertes-horrific-drug-war-philippines (Page Not Found)
http://www.thinkingpinoy.net/ (Dead Website)
http://thisisluv.com/ (Parked Domain)

REPLACED
- http://magdaloparty.com => https://magdalo.org/ (Updated URL)
- http://www.abs-cbnnews.com/ => https://news.abs-cbn.com/ (URL Redirect)
- http://efuckt.com/ => https://efukt.com/ (Updated URL)
- http://www.islamic-relief.org/category/where-we-work/philippines/ => https://islamic-relief.org/where_we_work/philippines/ (URL Redirect)
- https://www.c-r.org/where-we-work/southeast-asia/history-mindanao-conflict-and-moro-islamic-liberation-front/ => https://www.c-r.org/programme/southeast-asia/philippines-conflict-focus (URL Redirect)
- https://news.mb.com.ph/ => https://mb.com.ph/ (Updated URL)
- https://humanrightsinasean.info/members/philippines/ => https://hrasean.forum-asia.org/members/philippines/ (Updated URL)
[...]

PR #1620, opened 2024-01-23
Removed www.transparency.org duplicates in country-specific list

RandomGuy48:
#1461
> As the websites using https are encrypted, their content cannot be partly (e.g., separate pages) blocked. Thus, we aim to include only domains for such websites.


sloncocs:
Thanks so much, hugely appreciated 🙏

PR #1623, opened 2024-01-25
fix(www.jsf.mil): http => https

bassosimone:
As noted by https://github.com/ooni/probe/issues/2329, we cannot access http://www.jsf.mil/ but we can access https://www.jsf.mil/.

See also https://explorer.ooni.org/chart/mat?since=2023-12-26&until=2024-01-26&time_grain=day&axis_x=measurement_start_day&test_name=web_connectivity&domain=www.jsf.mil

Closes https://github.com/ooni/probe/issues/2329.


sloncocs:
LGTM, thanks @bassosimone! 🙏

PR #1683, opened 2024-03-14
Add TikTok to us.csv

c6670701:
Added website featured in a recent United State House of Representatives Bill. https://www.congress.gov/bill/118th-congress/house-bill/7521/text


sloncocs:
Thanks for your suggestion! 

TikTok is already being tested in the US as part of the Global test list. You can see the results of the tests here: https://explorer.ooni.org/chart/mat?probe_cc=US&since=2024-08-19&until=2024-09-19&time_grain=day&axis_x=measurement_start_day&test_name=web_connectivity&domain=www.tiktok.com

PR #1021, opened 2022-07-10
Added new URLs & updated ph.csv

RandomGuy48:
**ADDED:**
- https://liberation.ndfp.org/
- https://hiyaw.net/
- https://prwcinfo.wordpress.com/
- https://rctundfp.wordpress.com/
- https://compatriotsndf.wordpress.com/
- https://saveourschoolsnetwork.wordpress.com/
- https://umapilipinas.wordpress.com/
- https://ruralmissionaries.wordpress.com/
- https://rmp-national.weebly.com/
- https://pamalakayaweb.wordpress.com/
- https://angpamalakaya.org/
- https://amihanwomen.org/
- http://bagongalyansangmakabayan.org/
- https://ilps.info/
- https://www.counterpunch.org/
- https://iacenter.org/
- https://monthlyreview.org/
- https://peoples-march.blogspot.com/
- https://blog-by-taga-ilog-news.blogspot.com/
- https://partisan-news.blogspot.com/
- https://peopleresist.net/

These URLs were blocked last month but are now accessible [according to the news](https://youtu.be/kSHzPiZYsWw). However, that may not be the [case](https://www.bulatlat.com/2022/07/08/bulatlat-seeks-to-nullify-ntc-order-on-blocking-of-website/) yet.

image

_Resolutions that were present on the letter:_
1. https://atc.gov.ph/wp-content/uploads/2020/12/ATC-RESOLUTION-NO.-12-2020.pdf
2. https://atc.gov.ph/wp-content/uploads/2021/05/ATC-RESOLUTION-NO.-17-2021.pdf
3. https://atc.gov.ph/wp-content/uploads/2021/07/SGD_ATC-Resolution-No.-21_Designating-NDF.pdf

image
image

**UPDATED LINKS**
Removed the following dead URLs:
- http://www.prolife.org.ph/
- https://www.facebook.com/cenvisfardec
- https://www.proudcampaign.ph/

Updated the categories, description & added https for some URLs

PR #1686, opened 2024-03-23
Updated ph.csv

RandomGuy48:
**Removed:**
- http://www.pinoyexchange.com/ _(https://www.rappler.com/technology/internet-culture/pinoyexchange-website-closure-april-15-2023/)_
- http://pinoyhr.net/ _(empty WordPress site)_
- https://cnnphilippines.com/ _(https://www.sunstar.com.ph/tacloban/filipino-journalists-devastated-anew-after-cnn-philippines-shutdown)_
- http://lgbt.ie/ _(national organization in Ireland, irrelevant to ph.csv)_

**Reclassified into CTRL for being inactive**
- http://home.casema.nl/
- http://home.casema.nl/ndf/
- http://buhaymanggagawa.podomatic.com/
- https://bayan.ph/


sloncocs:
@RandomGuy48 do you think we should keep `http://home.casema.nl/` and `http://home.casema.nl/ndf/` in the list? they have been inactive since 2016, I'd suggest removing these URLs

sloncocs:
same for `https://bayan.ph/` as it hasn't been updated since 2014

RandomGuy48:
Yes, I agree. Both http://home.casema.nl/ & http://home.casema.nl/ndf/ seems to be gone.

I also removed other inactive websites that I labeled CTRL last year from https://github.com/citizenlab/test-lists/pull/1021/commits/03578a35ff7d9a691c7c0710d6879407d431147b
image

I also removed http://www.ihf-hr.org/. It has been previously removed from PR #1614 but has resurfaced again.

Lastly, I would like to ask about where to store the URLs from #1461 
> But at the same time, I think this list of URLs still can be useful for different purposes. So perhaps it would be useful to store these URLs somewhere else before removing them from testing.

sloncocs:
Thanks so much, @RandomGuy48! Regarding the URLs from #1461, unfortunately, we still have no consensus about where to move them. I would say unless there is a huge number of them and they are clearly affecting the number of measurements we receive for other websites, we keep them for the time being.

PR #1699, opened 2024-04-04
Update iMAP test lists

sitinurliza95:
Updated KH test list based on input from API Cambodia


sloncocs:
Hi @sitinurliza95! These websites seem to be still active to me, should we remove them?

http://www.dap-news.com/
https://www.licadho-cambodia.org/
https://cambodianess.com/

sitinurliza95:
@sloncocs Hi! Sorry I didn't notice your comment, yes we can keep those links for KH :)

PR #1703, opened 2024-04-10
Update br.csv

rathanaPhon:


sloncocs:
Hi @rathanaPhon! Thanks for your contribution! As far as I see, the `https:/www.chrisrathanajsorg.com/` resolves in NXDOMAIN error, so I would avoid adding it to the test list. I will close this PR for now, but let me know if you think it needs to be reopened.

PR #1702, opened 2024-04-10
DRL FAA IR Test List Update

sloncocs:
This pull request includes an extensive test list update to the Iranian test list by Miaan Group , as part of collaborative research on internet censorship in Iran.

PR #1726, opened 2024-05-13
DRL FAA IR Test List Update 

sloncocs:
This is a follow-up PR to #1702 adding more regional URLs.


agrabeli (approved):
Huge thanks to Miaan Group for contributing more updates to the test list for Iran. This is great work, thank you! <3

I think this PR is ready to be merged. Thanks again for this great contribution!

PR #1729, opened 2024-05-18
Removed https://www.webs.com/ and other duplicates on the global list

RandomGuy48:
image

Also replaced https://wordpress.org/openverse/ with https://openverse.org/


sloncocs:
thanks so much @RandomGuy48 🙏

PR #1904, opened 2025-01-28
Update mx.csv

carlosm2:
I removed some websites that hadn't been updated in over 10 years, as well as a couple that no longer exist

And  added 2 news websites and 2 websites that document issues related to missing persons.


sloncocs:
Thanks, @carlosm2! <3

PR #1909, opened 2025-02-05
Update sv.csv

carlosm2:
* add el faro news,
* add apes news collective,
* add agenciapresentes - lgbtq+ rights
* add CPJ com to protect journalists
 * add RSF report on SV


sloncocs:
@carlosm2 thanks for the contribution! 

I removed CPJ and RSF since both websites are on `https` it wouldn't be possible to block specific pages (or reports) hosted on the website, and the main domains are already being tested as part of the global test list. 

The rest looks good to me, merging it now 👍

carlosm2:
got it, thanks for the clarification about CPJ and RSF

PR #1925, opened 2025-03-27
Update Global test list

agrabeli:
This PR provides some updates to the Global test list, adding a few AI tool sites and other sites worth testing globally.


agrabeli:
I'll merge this so that we can start getting measurements globally for Amnezia VPN, which we'd like to feature in the Circumvention thematic page on OONI Explorer that we're launching soon.

PR #1936, opened 2025-04-11
Added Regavim

pyllyukko:
https://en.wikipedia.org/wiki/Regavim_(NGO)


sloncocs:
Thanks for your contribution @pyllyukko! Are there any specific countries where you think Regavim's website is blocked? Can we add it to some country-specific test lists rather than adding it to the global one?

pyllyukko:
> Thanks for your contribution @pyllyukko! Are there any specific countries where you think Regavim's website is blocked? Can we add it to some country-specific test lists rather than adding it to the global one?

Apparently at least in Finland. See: <https://x.com/KatriinaRepone1/status/1909949219256820014>

sloncocs:
> > Thanks for your contribution @pyllyukko! Are there any specific countries where you think Regavim's website is blocked? Can we add it to some country-specific test lists rather than adding it to the global one?
> 
> Apparently at least in Finland. See: https://x.com/KatriinaRepone1/status/1909949219256820014

Thanks! I removed it from the Global test list and added to Finland for now, if we see reports of it being unavailable in some other countries, we will add it to more test lists. 

We are avoiding adding more URLs to the Global test list unless we have reports that websites are blocked in multiple countries/regions. Global test list already exceeds 1500 URLs and is being tested in all countries all over the world, including countries with low connectivity and costly traffic. Adding more URLs to this list means that likely some of the URLs won't be tested at all in some countries and we could be missing some censorship events important for the local context (e.g., local human right websites or local political websites).

PR #1938, opened 2025-04-15
Remove squatted domain g30s-pki.com

Hans5958:
Some gambling site squatted this domain, so I think it is not relevant anymore.


sloncocs:
Thanks so much! Merging the PR now.

PR #1957, opened 2025-05-13
Update jo.csv

jifnawi:
Based on a leaked block order


sloncocs:
Thanks @jifnawi! I removed `www` from several websites since they redirect to the domains without `www`, merging it now 🙏

PR #1961, opened 2025-05-15
Remove cf ipfs gateways

hellais:
Cloudflare IPFS gateways are discontinued: https://blog.cloudflare.com/cloudflares-public-ipfs-gateways-and-supporting-interplanetary-shipyard/


hellais:
I am merging these as they are contributing to unnecessary measurements for all probes.

PR #1974, opened 2025-06-17
Create a directory for social media URLs

sloncocs:
This PR adds a new directory to store the URLs of specific social media pages and profiles. 

Currently, numerous URLs pertaining to specific profiles on Facebook, Twitter, YouTube, etc, are stored as part of the test lists, creating duplicates with the Global list. 

Since all social media are already tested globally, we suggest moving them to separate files, so that researchers could use them if needed. 

As part of the PR, I created an example csv for the UAE test list. I suggest keeping the same format of the tables and the same categorization to store social media URLs to make the lists more usable for the researchers in the future.


hellais:
Thanks for this PR, let's use this branch to collect social media URLs, but not merge it until we reach consensus on a reasonable format/structure for them.

This should not block on cleaning up the rest of the testing lists from social media URLs as it's all happening in a separate branch and directory.

PR #2127, opened 2026-02-20
th: Add more Thai government websites

bact:
- Update www.mfa.go.th and www.oic.go.th to HTTPS
- Add 35 new URLs, all government websites
  - All ministries
  - Common online services: Revenue, Social Security, election registration

Since around 2019, there are more reports that Thai websites in general can't be accessed from outside Thailand. This is coincide with GDPR enforcement and COVID-19.

Websites from agencies under the Ministry of Public Health are the most noticeable, also the government websites in general, but there are few other commercial websites too.

New URLs are extracted from this OONI Run list https://run.ooni.org/v2/10214 which is already expired, and the new/similar OONI Run list is created here: https://run.ooni.org/v2/10298


agrabeli:
@bact huge thanks for the contributions, and apologies for the delayed review! Merging now.

bact:
No worries! With the situation of many regions of the world today, OONI team can't be more busy. 🌞🏞️🌊

PR #2154, opened 2026-04-23
Update the current parties and government websites in pt.csv

joaopalmeiro:
Hi! 👋 

This PR updates the links in the `pt.csv` file to reflect the current political parties (plus two additional government websites).

Also, let me know if the source is what you expected or if I should update it to another value, please.

Thanks!

## References

- https://www.cne.pt/content/partidos-politicos-1
- https://www.parlamento.pt/sites/en/
- https://www.sg.mai.gov.pt/
- https://pt.wikipedia.org/wiki/Partido_Popular_Mon%C3%A1rquico and https://en.wikipedia.org/wiki/People%27s_Monarchist_Party_(Portugal)


hellais:
Thanks you for this contribution and apologies for taking so long to review it.

PR #2194, opened 2026-06-15
Added domains blocked by the CUII to de.csv

lina-x64:
These are domains that have been ordered to be blocked by the [CUII](https://cuii.info/), which all major German ISPs follow.
They have been compiled by us on [cuiiliste.de](https://cuiiliste.de/), here they have been de-duplicated, properly classified and removed if they appeared in other test lists.


hellais:
Thank you for suggesting the testing of these URLs. Merging.

PR #2195, opened 2026-06-15
Add domains blocked by German media authorities

lina-x64:
Added domains that are currently under blocking orders by German state media authorities. These were acquired through multiple transparency and press requests, which can be found here: https://fragdenstaat.de/projekt/liste-von-gesperrten-domains-05-05-2026


hellais:
Thank you for suggesting the addition of these URLs.

PR #2204, opened 2026-06-26
Add RT/Sputnik EU ban-evasion mirror domains

DebunkOrg:
This PR adds 112 RT and Sputnik domains — mostly EU ban-evasion mirror/clone sites that are currently unmonitored (zero OONI measurements). Each was content-verified (renders RT/Sputnik), and clones were confirmed via shared Yandex Metrika IDs and co-located hosting (89.191.237.0/24, 185.79.236.0/24). Goal: bring the evasion-mirror layer into OONI measurement so blocking gaps across EU countries become visible. Cross-referenced against the NLconnect EU media-sanctions reference list; ~12 are not on any public list.


hellais:
Thank you for contributing to the test lists.

Do you think it's going to be possible to cut down this list a bit more and move some of them, where relevant, to the country specific lists? The issue with adding so many URLs to the global list is that it's going to impact the coverage of the rest of the URLs inside of the global list which we try to maintain having as wide of a scope as possible.

I would suggest, if you wish to carry out this specific kind of very vertical study, that you make use of [OONI Run](https://run.ooni.org/) with a community of volunteers which you get to subscribe to your run link so that you can ensure adequate coverage.

In global list we could maybe include a few URLs from what you proposed in here.

PR #2231, opened 2026-07-30
Removed faulty URLs from Global test list

agrabeli:
As part of testing and checking the alerts generated by our prototype Censorship Alert System, we noticed several faulty URLs which constantly trigger false positives. This PR removes them.

Specifically, this PR implements the following:

* `valence.community` -> Removed because it's an expired domain
* `triller.co` -> Removed because it's an expired domain
* `ebuddy.com` -> Removed because it's no longer functional and often triggers false positives
* `badoo.com` -> Recategorized this entry from `GRP` to `DATE`


agrabeli:
This has been reviewed by @hellais , so I'll merge the PR.

PR #2248, opened 2026-08-07
Updates to ZM test list

agrabeli:
This PR updates the ZM test list based on:

* URLs shared by an OONI community member
* URLs from an OONI Run link used by local communities

This PR helps update the Zambian test list ahead of their 2026 elections next week.


agrabeli:
To help ensure we get as many measurements as possible ahead of the elections next week, I'll proceed with merging this PR.

PR #2258, opened 2026-08-27
Add Kiwix

### LINKED ISSUE #174: Implement censorship monitoring via OONI
We are now effectively victim of state censorship in Russia.

We have discovered it because users started to complain seriously.

We needed to run a few investigation to confirm this. See https://github.com/kiwix/operations/issues/656 for more information.

We could maybe use OONI to be informed earlier and in a rigourous manner about when, where and how our services get censored.

marek22k:
Closes https://github.com/kiwix/overview/issues/174


kelson42:
@marek22k Thank you very much for opening this PR. This would be really useful for Kiwix as we suffer of censorship meanwhile at least in one country... and this will probably become worse in the future. I'm here to answer any question if necessary.

